PR #7, opened 2016-10-24
Add `fr` unit and fix key order

lahmatiy:


SebastianZ:
Thank you for adding the unit! I missed it when I updated the CSS Grid property syntaxes.

Sebastian

SebastianZ:
According to this I've now [added the `<flex>` data type](https://github.com/mdn/data/commit/16fbfee6b3ea376c872f26c19c099df7e12884eb).

Sebastian

PR #9, opened 2016-10-24
Replace deprecated <uri> for <url>

lahmatiy:
> ... CSS Level 3 came back to the initial definition, this time explicitly defined. The functional notation url() denotes a <url> CSS data type and no more the more generic <uri> CSS data type.

https://developer.mozilla.org/en-US/docs/Web/CSS/url

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -46,7 +46,7 @@
     "status": "nonstandard"
   },
   "-moz-binding": {
-    "syntax": "&lt;uri&gt; | none",
+    "syntax": "&lt;url&gt; | none",
     "media": "visual",
     "inherited": false,
     "animationType": "discrete",
@@ -2792,7 +2792,7 @@
     "status": "experimental"
   },
   "content": {
-    "syntax": "normal | none | [ &lt;string&gt; | &lt;uri&gt; | &lt;counter&gt; | attr() | open-quote | close-quote | no-open-quote | no-close-quote ]+",
+    "syntax": "normal | none | [ &lt;string&gt; | &lt;url&gt; | &lt;counter&gt; | attr() | open-quote | close-quote | no-open-quote | no-close-quote ]+",
     "media": "all",
     "inherited": false,
     "animationType": "discrete",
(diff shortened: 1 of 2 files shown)


teoli2003:
Some of these uri -> url has not made it to the spec [some will not directly be changed, because the syntax will be extended in the future (list-style-image will use <image> in the future)

But I think we should go ahead with these changes.

css/properties.json
```diff
@@ -2792,7 +2792,7 @@
     "status": "experimental"
   },
   "content": {
-    "syntax": "normal | none | [ &lt;string&gt; | &lt;uri&gt; | &lt;counter&gt; | attr() | open-quote | close-quote | no-open-quote | no-close-quote ]+",
+    "syntax": "normal | none | [ &lt;string&gt; | &lt;url&gt; | &lt;counter&gt; | attr() | open-quote | close-quote | no-open-quote | no-close-quote ]+",
```
SebastianZ: FYI, CSS 2.2 still [defines it using `<uri>`](https://www.w3.org/TR/CSS22/generate.html#propdef-content), but [it was resolved](https://lists.w3.org/Archives/Public/www-style/2016Oct/0091.html) to refer to [CSS Syntax](https://drafts.csswg.org/css-syntax/) instead, which removed `<uri>`. So it's fine that it gets changed already.

Sebastian

css/properties.json
```diff
@@ -4074,7 +4074,7 @@
     "status": "standard"
   },
   "list-style-image": {
-    "syntax": "&lt;uri&gt; | none",
+    "syntax": "&lt;url&gt; | none",
```
SebastianZ: Same as for `content`.

Sebastian

SebastianZ:
> Some of these uri -> url has not made it to the spec [some will not directly be changed, because the syntax will be extended in the future (list-style-image will use in the future)

I've provided some more details in my review. The W3C already [resolved on changing CSS 2.2 property syntaxes to refer to CSS Syntax](https://github.com/w3c/csswg-drafts/issues/412). See https://github.com/w3c/csswg-drafts/issues/412.

Sebastian

PR #12, opened 2016-10-27
Fixed <'text-indent'> syntax according to spec

lahmatiy:
https://drafts.csswg.org/css-text-3/#text-indent-property


SebastianZ:
Thank you for the patch!

Sebastian

PR #16, opened 2016-10-28
Update <'content'> syntax and related changes

lahmatiy:
Last spec (https://drafts.csswg.org/css-content/) has problems with syntaxes. Most of them addressed in related issue https://github.com/w3c/csswg-drafts/issues/663
Syntaxes added with fixes.
Probably we shouldn't add `<target>` and `<leader>` syntaxes until spec stability.


teoli2003:
There are some comments by @SebastianZ  on w3c/csswg-drafts#663 

I defer to him for the review here.

SebastianZ:
Sorry for the long delay! The changes look good to me and they refer to what the spec. says, so I've merged them.

Though we should be aware that no browser actually implements the whole new syntax yet, that we need to update the [MDN page of the `content` property](https://developer.mozilla.org/en-US/docs/Web/CSS/content) accordingly and that the syntax is likely to change in the future, so we'll probably need to update it again.

Sebastian

PR #18, opened 2016-11-02
change scroll-snap-points-* to use length-percentage

Manishearth:
Only firefox implements these, and it's deprecated (and removed from the then-draft spec) so no other browser is likely to start implementing this. We should just specify what firefox does in this case (which is handle both lengths and percentages)

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5366,7 +5366,7 @@
     "status": "standard"
   },
   "scroll-snap-points-x": {
-    "syntax": "none | repeat( &lt;length&gt; )",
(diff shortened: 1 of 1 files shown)


teoli2003:
Yep, both are indeed working.

SebastianZ:
For reference, I've now updated the MDN descriptions of [`scroll-snap-points-x`](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-points-x) and [`scroll-snap-points-y`](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-points-y) accordingly.

Sebastian

PR #22, opened 2016-11-24
Fix webkit-text-stoke and webkit-text-fill-color properties

chenpighead:
According to the [CSS Compatibility Specification](https://compat.spec.whatwg.org/#the-webkit-text-stroke), fix some of these values.

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -670,15 +670,16 @@
     ],
     "initial": "<code>currentcolor</code>",
     "appliesto": "allElements",
-    "computed": "asSpecified",
+    "computed": "'color'",
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
   "-webkit-text-stroke": {
-    "syntax": "&lt;length&gt; &lt;color&gt;",
+    "syntax": "&lt;length&gt; || &lt;color&gt;",
     "media": "visual",
     "inherited": true,
     "animationType": [
+      "-webkit-text-stroke-width",
       "-webkit-text-stroke-color"
     ],
     "percentages": "no",
@@ -690,8 +691,11 @@
       "-webkit-text-stroke-color"
     ],
     "appliesto": "allElements",
-    "computed": "asSpecified",
-    "order": "uniqueOrder",
+    "computed": [
+      "-webkit-text-stroke-width",
+      "-webkit-text-stroke-color"
+    ],
+    "order": "canonicalOrder",
     "status": "nonstandard"
   },
   "-webkit-text-stroke-color": {
@@ -703,9 +707,9 @@
     "groups": [
       "WebKit Extensions"
     ],
-    "initial": "<code>rgb(0, 0, 0)</code>",
+    "initial": "<code>currentColor</code>",
     "appliesto": "allElements",
-    "computed": "asSpecified",
+    "computed": "'color'",
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
@@ -720,7 +724,7 @@
     ],
     "initial": "<code>0</code>",
     "appliesto": "allElements",
-    "computed": "asSpecified",
+    "computed": "absoluteWidth",
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
--- a/css/properties.json
+++ b/css/properties.json
@@ -707,7 +707,7 @@
     "groups": [
       "WebKit Extensions"
     ],
-    "initial": "<code>currentColor</code>",
+    "initial": "<code>currentcolor</code>",
     "appliesto": "allElements",
     "computed": "'color'",
     "order": "uniqueOrder",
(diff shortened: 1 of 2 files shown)


css/properties.json
```diff
@@ -399,7 +399,7 @@
     "groups": [
       "WebKit Extensions"
     ],
-    "initial": "<code>currentcolor</code>",
+    "initial": "<code>currentColor</code>",
```
SebastianZ: Please keep `currentcolor` lowercase! This makes it consistent with other keywords and the spec.

Sebastian

css/properties.json
```diff
@@ -668,9 +668,9 @@
...
```
SebastianZ: Same as above.

Sebastian

css/properties.json
```diff
@@ -679,6 +679,7 @@
     "media": "visual",
```
SebastianZ: According to the spec. the syntax of `-webkit-text-stroke` actually also needs to be changed to `"&lt;line-width&gt; || &lt;color&gt;"`.

Sebastian

css/properties.json
```diff
@@ -690,8 +691,11 @@
       "-webkit-text-stroke-color"
     ],
     "appliesto": "allElements",
-    "computed": "asSpecified",
-    "order": "uniqueOrder",
+    "computed": [
+      "-webkit-text-stroke-width",
+      "-webkit-text-stroke-color"
+    ],
+    "order": "orderOfAppearance",
```
SebastianZ: Order of appearance seems wrong, because the color and width can appear in any order. Maybe "canonicalOrder" is better here.

Sebastian

css/properties.json
```diff
@@ -703,24 +707,24 @@
     "groups": [
       "WebKit Extensions"
     ],
-    "initial": "<code>rgb(0, 0, 0)</code>",
+    "initial": "<code>currentColor</code>",
```
SebastianZ: Again, `currentcolor`.

Sebastian

css/properties.json
```diff
@@ -703,24 +707,24 @@
...
+    "computed": "'color'",
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
   "-webkit-text-stroke-width": {
     "syntax": "&lt;length&gt;",
     "media": "visual",
     "inherited": true,
-    "animationType": "discrete",
+    "animationType": "no",
```
SebastianZ: Please keep `"discrete"`. The W3C [decided some time ago](https://lists.w3.org/Archives/Public/www-style/2015Nov/0308.html) to replace the "animatable" info by an "animation type" [with "discrete" as value](https://lists.w3.org/Archives/Public/www-style/2016May/0121.html) when they have a discrete value transition.

The spec. obviously didn't get updated in this regard yet.

Sebastian

css/properties.json
```diff
@@ -703,24 +707,24 @@
...
     "percentages": "no",
     "groups": [
       "WebKit Extensions"
     ],
     "initial": "<code>0</code>",
     "appliesto": "allElements",
-    "computed": "asSpecified",
+    "computed": "absolute width",
```
SebastianZ: This should be written as "absoluteWidth" (as l10n keyword) and get translated in [l10n/css.json](https://github.com/mdn/data/blob/master/l10n/css.json).

Sebastian

css/properties.json
```diff
@@ -1340,7 +1344,7 @@
     "groups": [
       "CSS Logical Properties"
     ],
-    "initial": "<code>currentcolor</code>",
+    "initial": "<code>currentColor</code>",
```
SebastianZ: Lowercase like above.

Sebastian

css/properties.json
```diff
@@ -1408,7 +1412,7 @@
...
```
SebastianZ: Here, too.

Sebastian

css/properties.json
```diff
@@ -1773,7 +1777,7 @@
...
```
SebastianZ: And here.

Sebastian

css/properties.json
```diff
@@ -1841,7 +1845,7 @@
...
```
SebastianZ: And also here.

Sebastian

css/properties.json
chenpighead: Does that mean we shall keep `currentcolor`  lowercase in all usecases?

css/properties.json
chenpighead: Thank you for the explanation. I shall file an issue to Web Compatibility Specification for updating this.

css/properties.json
chenpighead: Don't know we have such a keyword. Will fix this. Thank you!

css/properties.json
chenpighead: Filed https://github.com/whatwg/compat/issues/66 for this.

css/properties.json
chenpighead: Good catch. Will fix it.

css/properties.json
SebastianZ: > Don't know we have such a keyword.

Looks like this is not fixed yet. This keyword doesn't exist yet. You need to add it to css/properties.json and you need to add an entry for it in the translation file l10n/css.json. If you have any questions about that, please let me know.

Sebastian

chenpighead: Oops, missed this one. Fixed this in the latest push.

PR #26, opened 2016-12-06
Use absoluteLength instead of absoluteWidth for -webkit-text-stroke-width's computed value

chenpighead:
According to the spec, '-webkit-text-stroke-width' property uses <line-width> as
its specified value. And all the <line-width>s properties seem to be computed to
'absolute length', not 'absolute width'. So, let's use 'absolute length' for
the computed value of '-webkit-text-stroke-width' property.

Spec issue: https://github.com/whatwg/compat/issues/67

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -724,7 +724,7 @@
     ],
     "initial": "<code>0</code>",
     "appliesto": "allElements",
-    "computed": "absoluteWidth",
+    "computed": "absoluteLength",
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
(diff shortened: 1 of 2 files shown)


SebastianZ:
Sorry for the delay! I was at a conference and just came back today.
The change looks good to me, though for some reason I don't have write access to the repo anymore to merge it.
@Elchi3, can you give me access again, please?

Sebastian

jwhitlock:
@SebastianZ that was me, I'll email you next steps to get access again.

@chenpighead thanks for your contribution!

chenpighead:
@SebastianZ thanks for the review.
@jwhitlock thanks for the help! :)

PR #27, opened 2016-12-15
user-select initial property should be 'auto'

maxkirchoff:
According to the draft spec on user-select that is sourced from MDN docs, the initial for user-select is 'auto' and not 'none'. This is also true in the way it is implemented in supported browsers.
https://drafts.csswg.org/css-ui-4/#propdef-user-select

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6075,7 +6075,7 @@
     "groups": [
       "Mozilla Extensions"
     ],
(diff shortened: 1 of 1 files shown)


SebastianZ:
Merged the change. Thank you!

I assume, the initial value was different in previous spec. versions.

Sebastian

PR #28, opened 2016-12-20
Add pinch-zoom into the touch-action css property.

dtapuska:
See https://github.com/whatwg/compat/issues/68

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5890,7 +5890,7 @@
     "status": "standard"
   },
   "touch-action": {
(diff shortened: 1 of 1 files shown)


dtapuska:
@RByers @teddink @navidz @mustaqahmed @jpmedley  This adds it to the MDN CSS syntax page...

RByers:
Thanks LGTM.  This is now spec'd officially [here](https://compat.spec.whatwg.org/#touch-action) - can you also update MDN to link to that spec?

SebastianZ:
Merged the patch. Thanks!

> can you also update MDN to link to that spec?

[Done](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action).

Sebastian

PR #33, opened 2017-01-18
Add "auto" to -moz-user-input

upsuper:
From [the code](https://dxr.mozilla.org/mozilla-central/rev/6a23526fe5168087d7e4132c0705aefcaed5f571/layout/style/nsRuleNode.cpp#5273-5278), it seems the initial value is actually "auto" rather than "none".


SebastianZ:
You're right, Xidorn. Thanks for your input!

Sebastian

PR #38, opened 2017-01-27
Revert "adding percentage as an option in alpha-value, angle as an option in hue, and alpha-value as an optional fourth parameter in hsl() and rgb(), as per CSS4 color"

SebastianZ:
This reverts commit b0211788972b3c4530b37a229ec872f9ee6ed207.


chrisdavidmills:
I can merge this, so I have done. Looks fine. Thanks for fixing this mess, Sebastian ;-)

PR #44, opened 2017-02-02
Update syntax of <mask-layer>

upsuper:
This was changed in w3c/fxtf-drafts#92.


SebastianZ:
Thanks for this PR and generally for helping with keeping the syntaxes updated!

Sebastian

PR #46, opened 2017-02-09
-moz-image-region can be auto

Manishearth:
Only Firefox handles this property, and we do parse auto values for it.


SebastianZ:
Thanks for the patch! `auto` is even the initial value.

Sebastian

PR #47, opened 2017-02-13
Correct capitalisation of currentColor keyword

lesleh:
As specified in [CSS Color Module Level 3](https://www.w3.org/TR/css3-color/#currentcolor).


SebastianZ:
All keywords in CSS are written lowercase and `currentcolor` was [changed accordingly in CSS Color 4](https://drafts.csswg.org/css-color/#currentcolor-color).

Sebastian

PR #48, opened 2017-02-14
Updating initial values of mask-position and mask-repeat

chrisdavidmills:
as per [bug 1308963](https://bugzilla.mozilla.org/show_bug.cgi?id=1308963)

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -4423,7 +4423,7 @@
     "groups": [
       "CSS Masks"
     ],
-    "initial": "<code>center</code>",
+    "initial": "<code>0%</code>",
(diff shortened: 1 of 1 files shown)


css/properties.json
```diff
@@ -4423,7 +4423,7 @@
     "groups": [
       "CSS Masks"
     ],
-    "initial": "<code>center</code>",
+    "initial": "<code>0%</code>",
```
SebastianZ: That's actually `0% 0%`, as it covers the x and y components.

Sebastian

chrisdavidmills:
Value updated. Thanks!

SebastianZ:
Thank you for the quick fix!

Sebastian

PR #49, opened 2017-02-16
adding font-variation-settings

chrisdavidmills:


css/properties.json
```diff
@@ -3240,6 +3240,26 @@
     ],
     "status": "standard"
   },
+  "font-variation-settings": {
+    "syntax": "normal | [ &lt;string&gt; &lt;number&gt; ] #",
```
SebastianZ: Nit: We're not adding spaces between `]` and the multiplier. Can you please remove it?

Sebastian

css/properties.json
```diff
@@ -3240,6 +3240,26 @@
...
+    "media": "visual",
+    "inherited": true,
+    "animationType": "discrete",
```
SebastianZ: The [spec. says](https://drafts.csswg.org/css-fonts-4/#low-level-font-variation-settings-control-the-font-variation-settings-property):

> Animating font-variation-settings is possible using a mechanism similar to animating the transform property: Two declarations of font-feature-settings can be animated between if they are "like".

So, `discrete` is wrong here. There needs to be a short description added to l10n/css.json explaining the animation type and a referencing keyword here.

Sebastian

css/properties.json
```diff
@@ -3240,6 +3240,26 @@
...
+    "percentages": "no",
+    "groups": [
+      "CSS Fonts"
+    ],
+    "initial": "<code>normal</code>",
+    "appliesto": "allElements",
+    "computed": "asSpecified",
+    "order": "uniqueOrder",
```
SebastianZ: Please use `perGrammar` here. (It's actually the same, though it's what the spec. says.)

Sebastian

css/properties.json
```diff
@@ -3240,6 +3240,26 @@
...
+    "alsoAppliesTo": [
+      "::first-letter",
+      "::first-line",
+      "::placeholder"
+    ],
+    "status": "standard"
```
SebastianZ: As the spec. is still an Editor's Draft, I think this should be marked as `experimental`, even when there are already multiple implementations.

Sebastian

chrisdavidmills:
I've attempted to update the data as per your comments. Let me know if this is ok. Cheers for the review!

PR #50, opened 2017-02-20
make max-inline-size and max-block-size experimental

iamstarkov:
somehow these props were not marked as experimental, i fixed it


SebastianZ:
Thanks for the patch, Vladimir!

Sebastian

PR #57, opened 2017-04-04
Add missed # for <color-stop-list>

lahmatiy:
Color stops should be separated by comma.

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -27,7 +27,7 @@
   "clip-source": "<url>",
   "color": "<rgb()> | <rgba()> | <hsl()> | <hsla()> | <hex-color> | <named-color> | currentcolor | <deprecated-system-color>",
   "color-stop": "<color> <length-percentage>?",
-  "color-stop-list": "<color-stop>{2,}",
+  "color-stop-list": "<color-stop>#{2,}",
   "common-lig-values": "[ common-ligatures | no-common-ligatures ]",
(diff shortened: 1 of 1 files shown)


css/syntaxes.json
```diff
@@ -27,7 +27,7 @@
   "clip-source": "<url>",
   "color": "<rgb()> | <rgba()> | <hsl()> | <hsla()> | <hex-color> | <named-color> | currentcolor | <deprecated-system-color>",
   "color-stop": "<color> <length-percentage>?",
-  "color-stop-list": "<color-stop>{2,}",
+  "color-stop-list": "<color-stop>#{2,}",
```
jwhitlock: The css-images-3 spec doesn't include the ``#``: https://drafts.csswg.org/css-images-3/#color-stop-syntax.

The css-images-4 spec has the ``#``, but is more complex:
https://drafts.csswg.org/css-images-4/#color-stop-syntax

I suspect the css-images-3 spec is wrong, but I'm not sure what we do in these situations.

a2sheppy: Since the “#” doesn’t break compatibility with actual code and examples from the -3 spec, and because it clarifies the real-world syntax, I think this is a good change.

lahmatiy: It's a mistake in css-image-3, because color-stop-list with no comma doesn't work in browsers and all examples in spec contain a comma.
I filled [an issue](https://github.com/w3c/csswg-drafts/issues/458) about mistake about half year ago, but no reaction yet.

lahmatiy: Issue has been resolved. Latest version css-images-3 spec includes `#` in `<color-stop-list>` grammar.

PR #75, opened 2017-05-12
Fix some "appliesto" values

lahmatiy:
- `border-image`
    - the single property with an array value, changed to "allElementsExceptTableElementsWhenCollapse" as its longhand properties
- `border-image-width`
    - `allElementsExceptTableElementsWhenBorderCollapseCollapse` -> `allElementsExceptTableElementsWhenCollapse` like other `border-image-*` properties; The single difference is missed `internal` word in first description. I guess it's [a mistake in spec](https://drafts.csswg.org/css-backgrounds-3/#the-border-image-width) because there is no reasons to differ from other `border-image-*` properties.
- `mix-blend-mode`
    - `allHTMLElements` -> `allElements` (see https://drafts.fxtf.org/compositing-1/#mix-blend-mode)

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -1635,13 +1635,7 @@
       "border-image-outset",
       "border-image-repeat"
     ],
-    "appliesto": [
-      "border-image-outset",
-      "border-image-repeat",
-      "border-image-slice",
-      "border-image-source",
-      "border-image-width"
-    ],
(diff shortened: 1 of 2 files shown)


lahmatiy:
Done!
I've created a PR to spec to fix an issue with "internal": https://github.com/w3c/csswg-drafts/pull/1379

PR #84, opened 2017-05-25
adding moz-context-properties

chrisdavidmills:
I am really not sure what values to put here, so I'm taking my best guess, with the intention of then asking the engineer if he can verify these values.

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -121,7 +121,7 @@
     "status": "nonstandard"
   },
   "-moz-context-properties": {
-    "syntax": "fill || stroke",
+    "syntax": "none | [ fill | stroke ]#",
     "media": "visual",
     "inherited": false,
     "animationType": "discrete",
--- a/css/properties.json
+++ b/css/properties.json
@@ -120,6 +120,21 @@
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
+  "-moz-context-properties": {
+    "syntax": "fill || stroke",
+    "media": "visual",
+    "inherited": false,
+    "animationType": "discrete",
+    "percentages": "no",
+    "groups": [
+      "Mozilla Extensions"
+    ],
+    "initial": "none",
+    "appliesto": "replacedElements",
+    "computed": "asSpecified",
+    "order": "uniqueOrder",
+    "status": "nonstandard"
+  },
   "-moz-float-edge": {
     "syntax": "border-box | content-box | margin-box | padding-box",
     "media": "visual",


css/properties.json
```diff
@@ -120,6 +120,21 @@
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
+  "-moz-context-properties": {
+    "syntax": "fill || stroke",
```
lahmatiy: Should be `[ fill | stroke ]#`

chrisdavidmills: Thanks for the input. I was confused about this. It can accept either, or both separated by a comma, so

fill
stroke
fill, stroke

Given that, is ```[ fill | stroke ]#``` still correct?

lahmatiy: Yep, see https://www.w3.org/TR/css3-values/
`fill | stroke` - enum
`[]#` - group occurs one or more times, separated by comma

Correction: property can contain `none` value, so syntax should be `none | [ fill | stroke]#`

chrisdavidmills: Cool - thanks a lot for your help. I have corrected and resubmitted it

css/properties.json
```diff
@@ -120,6 +120,21 @@
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
+  "-moz-context-properties": {
+    "syntax": "none | [ fill | stroke ]#",
```
jwatt: You can add the values fill-opacity and stroke-opacity now.

css/properties.json
```diff
@@ -120,6 +120,21 @@
...
+    "media": "visual",
+    "inherited": false,
```
jwatt: This should be true.

css/properties.json
```diff
@@ -120,6 +120,21 @@
...
+    "animationType": "discrete",
+    "percentages": "no",
+    "groups": [
+      "Mozilla Extensions"
+    ],
+    "initial": "none",
+    "appliesto": "replacedElements",
```
jwatt: It doesn't just apply to replaced elements. It applies to any element that can reference an image.

jwatt:
I made some comments on the patch. Not sure if you're notified of them automatically though.

chrisdavidmills:
Thanks jwatt! I made the changes in a separate PR; see:

https://github.com/mdn/data/pull/91

One comment — I've set "appliesto" as  "blockLevelElements", as I'm not 100% sure what to set this as. I don't think it's exactly right, but it is closer than "replacedElements". ;-) I'll have to research what this needs to be changed to.

PR #99, opened 2017-06-30
Added proper groups schema validation

SebastianZ:
Added a detailed groups validation to the JSON schema, split it out to a separate file and fixed/updated the groups within the data accordingly.


css/properties.schema.json
```diff
@@ -12,12 +12,53 @@
         }
       }
     },
-    "stringList": {
+    "groupList": {
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
       "items": {
```
lahmatiy: Should be `"$ref": "definitions.json#/groupList"`?

lahmatiy:
It's awesome! Would be cool to add other enums to `definitions.json`

SebastianZ:
> It's awesome! Would be cool to add other enums to `definitions.json`

Yes, that's the goal. This change is just the first step.

Sebastian

css/properties.schema.json
SebastianZ: Thanks for pointing that out! This got actually redundant, because we have the data in definitions.json. So I just removed this item in https://github.com/mdn/data/pull/99/commits/a2f237368dfd06d0b9501c825a61a442b98fd6ce.

Sebastian

PR #106, opened 2017-09-08
Remove proprietary pseudo-elements from the list.

mfluehr:
This list is far from including all the proprietary pseudo-elements, but if it did the list of pseudo-elements would be unnecessarily long. I think we should let the https://developer.mozilla.org/en-US/docs/Web/CSS/Mozilla_Extensions, https://developer.mozilla.org/en-US/docs/Web/CSS/Webkit_Extensions, and https://developer.mozilla.org/en-US/docs/Web/CSS/Microsoft_CSS_extensions pages cover it.


mfluehr:
If we don't want to actually remove these, is there a way to hide them in the pseudo-elements list?

PR #113, opened 2017-09-25
Add `hanging-punctuation`

mfluehr:


lahmatiy:
There is no `inlineElements` in `properties.schema.json` for `appliesto` (checking failed). I'm not sure that spec is correct here since value is never used for properties before. On the other hand, it may be correct and that's the first time when such value is needed. In last case a new value for `appliesto` should be added to schema.

lahmatiy:
Reminder: page on MDN should be updated (https://developer.mozilla.org/en-US/docs/Web/CSS/hanging-punctuation) after the changes merge

mfluehr:
Ooh, that's a tricky question about `inlineElements`. What would you recommend? Maybe just list it as "all elements" until the spec gets settled? I have to admit, my comprehension of "Applies to" is fairly limited.

lahmatiy:
Yep, I think it should be "all elements". We can use this value for now and create an issue with question in [spec bug tacker](https://github.com/w3c/csswg-drafts/labels/css-text-3)

mfluehr:
@lahmatiy Ok, the checks are passing now, thanks.

mfluehr:
Spec bug for the correct "applies to" value created here: https://github.com/w3c/csswg-drafts/issues/1833

PR #114, opened 2017-09-25
Fix `symbol` syntax

mfluehr:
This is used in the `symbols` descriptor, part of the `@counter-style` at-rule.


teoli2003:
https://drafts.csswg.org/css-counter-styles/#symbols-function doesn't have it. Which spec makes use of this?

mfluehr:
It's found in https://drafts.csswg.org/css-counter-styles-3/#counter-style-symbols .

PR #115, opened 2017-10-04
Add `shape-box` data type

mfluehr:
This data type is used in the `shape-outside` property.

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -549,7 +549,7 @@
     "syntax": "step-start | step-end | steps(<integer>[, [ start | end ] ]?)"
   },
   "symbol": {
-    "syntax": "<string> | <image> | <ident>"
+    "syntax": "<string> | <image> | <custom-ident>"
   },
   "target": {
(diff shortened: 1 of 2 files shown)


teoli2003:
I disagree on this one. <shape-box> is not a fundamental CSS types used everyone. It is a syntactical type used for technical reasons and defined in syntaxes.json L512.

I would not make a specific article for it, I would simply list all the keywords and their meaning in shape-outside.

mfluehr:
Thanks for the response, @teoli2003 . The article already exists (created in 2014): https://developer.mozilla.org/en-US/docs/Web/CSS/shape-box . Do you want to delete it, and remove the link to it found in https://developer.mozilla.org/en-US/docs/Web/CSS/shape-outside ?

teoli2003:
Thank for your help, mfluehr!

I think we should:
1) Integrate the info from the shape-box page into shape-outside (no <shape-box> should appear under the sections 'Values'
2) Redirect the shape-box page to shape-outside.

mfluehr:
Alright, I've updated those two pages per your recommendations. Closing this pull request.

PR #129, opened 2017-10-26
Move `fr` unit to "CSS Flexible Lengths"

mfluehr:
The link for `fr` on https://developer.mozilla.org/en-US/docs/Web/CSS/Reference points to the wrong page. This is because the `fr` unit doesn't belong with the other lengths, such as `em` or `px`. Rather, it is defined on the `<flex>` data type page.

We will also need to update https://github.com/mdn/kumascript/blob/master/macros/CSS_Ref.ejs.

### DIFF
--- a/css/definitions.json
+++ b/css/definitions.json
@@ -18,6 +18,7 @@
       "CSS Device Adaptation",
       "CSS Display",
       "CSS Flexible Box Layout",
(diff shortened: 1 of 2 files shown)


mfluehr:
It seems perhaps a little ugly to create a new group just for this, so I'm open to alternative suggestions.

PR #150, opened 2017-11-21
updating text-decorsation-skip, and adding text-decoration-skip-ink

chrisdavidmills:


toptalo:
Hello teoli2003, chrisdavidmills
I think it's too early to use this formal syntax.
Specification is only has ED status. But there are CSS Text Decoration Module Level 3 that has CR status

teoli2003:
Hi!
The status of the spec is irrelevant. The criteria are the agreement between browser vendors and the implementation (unprefixed, not behind a pref) by a browser vendor.  In this case, the second criteria is fulfilled, and the issues in github and bugzilla seem to indicate that their is a consensus around it (I've not double checked the W3C meeting notes, though).

Hope this helps.

toptalo:
But current browsers even level 3 do not fully support

https://caniuse.com/#search=text-decoration

toptalo:
And there are some confusion in this article https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip

See section `Syntax` and `Formal Syntax`

chrisdavidmills:
@toptalo Can you let me know what is wrong in the Syntax/Formal Syntax sections? It took me a while to figure out how to fill these in ;-)

I don't think ```text-decoration-skip: spaces leading-spaces trailing-spaces;``` is wrong — you are allowed to put in both leading and trailing spaces?

toptalo:
1. There are `ink` value in Syntax section that are outdated.
2. Here are formal syntax `none | [ objects || [ spaces | [ leading-spaces || trailing-spaces ] ] || edges || box-decoration ]`, where
`|` - Separating two or more entities by a single bar, means that all entities are exclusive options: exactly one of these options must be present. This is typically used to separate a list of possible keywords.
`||` - Separating two or more components by a double bar, means that all entities are options: at least one of them must be present, and they may appear in any order. Typically this is used to define the different values of a shorthand property

So you can not both spaces and leading-spaces trailing-spaces in one time.

chrisdavidmills:
Ah yes, I get where the problem is now — I understand what the single/double bar syntax means in general, but this one got me really confused.

Thanks for your help ;-)

PR #152, opened 2017-11-23
Fix AudioContext & OfflineAudioContext inheritance

Jedipedia:
[`AudioContext`](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext) and [`OfflineAudioContext`](https://developer.mozilla.org/en-US/docs/Web/API/OfflineAudioContext) both inherit from [`BaseAudioContext`](https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext).

See also the [Web Audio API spec](https://www.w3.org/TR/webaudio/#idl-def-AudioContext).

### DIFF
--- a/api/inheritance.json
+++ b/api/inheritance.json
@@ -1378,7 +1378,7 @@
     ]
   },
   "OfflineAudioContext": {
-    "inherits": "AudioContext",
+    "inherits": "BaseAudioContext",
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
I've reviewed this, and it looks fine. r+

PR #157, opened 2018-01-16
Bring data on specs extracted from kumascript

dontcallmedom:
The purpose of extracting it is to make it easier to validate and update it - I have a number of ideas and tools that I can apply to this, but it's much easier to apply them if the data is cleanly separated.

See also https://github.com/mdn/kumascript/pull/557


teoli2003:
I think we should talk together before reviewing this, this has consequences and has to be done in the right order. Meanwhile I think we should not review this.

Can you open an issue with the goal, or on discourse MDN?

ExE-Boss:
I’m&nbsp;now doing&nbsp;this in&nbsp;#397 using&nbsp;the&nbsp;unified spec&nbsp;data.

Rumyra:
Kumascript is now archived so closing

Issue #159, opened 2018-01-16
<bg-layer> and <final-layer> formal syntax differs from W3C


wbamberg:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1428255.

The data should probably match the W3C syntax.

lahmatiy:
Agree. It must be changed. The difference (from easiest):
- `<box>{1,2}` instead of `<box> || <box>` in `<bg-layer>`. As Russ Weakley (reporter) notes:
    > While this suggests the same thing as the W3C, it does differ.

    So this must be changed as incorrect.
- `<'background-color'>` position in `<final-bg-layer>` – should be in the beginning according to W3C spec. It doesn't change anything since color doesn't intersect with other part of value. But moving it to a right place allows to be aligned with W3C
- `<bg-position>` instead of `<position>`. It's complicated one. According to dictionaries `<position>` is used in following properties:
    - `background`
    - `background-position`
    - `mask`
    - `mask-position`
    - `object-position`
    - `offset`
    - `offset-anchor`
    - `offset-position`
    - `perspective-origin`
    - `scroll-snap-coordinate`
    - `scroll-snap-destination`

    [Old version](https://www.w3.org/TR/2014/CR-css3-background-20140909/#ltpositiongt) (Sept, 2014) of CSS Backgrounds and Borders Module Level 3 defined syntax with `<position>`. But [latest version](https://www.w3.org/TR/css-backgrounds-3/) uses `<bg-position>` with changed [definition](https://drafts.csswg.org/css-backgrounds/#typedef-bg-position). At the same time, [CSS Masking Module Level 1](https://drafts.fxtf.org/css-masking-1) and [Motion Path Module Level 1](https://drafts.fxtf.org/motion-1/) are using `<position>` (suppose specs for other properties too) that refers to [CSS Values and Units Module Level 4](https://drafts.csswg.org/css-values-4/#typedef-position) with its own [`<position>` definition](https://drafts.csswg.org/css-values-4/#position). And here begins the most interesting, CSS Values and Units Module Level 4 claims:
    > The `<position>` value specifies the position of a object area (e.g. background image) inside a positioning area (e.g. background positioning area). It is interpreted as specified for background-position.  [CSS3-BACKGROUND]

    And then:
    > Note: The background-position property also accepts a three-value syntax. This has been disallowed generically because it creates parsing ambiguities when combined with other length or percentage components in a property value.

[...]

PR #160, opened 2018-01-16
Change background syntaxes according to W3C spec

lahmatiy:
- Changed `<bg-layer>` and `<final-bg-layer>` per [CSS Backgrounds and Borders Module Level 3](https://www.w3.org/TR/css-backgrounds-3/)
- Renamed `<position>` to `<bg-position>`
- Added `<position>` according to [CSS Values and Units Module Level 4](https://drafts.csswg.org/css-values/#position)

See #159 for details

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -1231,7 +1231,7 @@
     "status": "standard"
   },
   "background-position": {
-    "syntax": "<position>#",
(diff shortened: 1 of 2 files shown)


wbamberg:
Tested on a local Kuma with the `background` page, looks fine.

PR #162, opened 2018-01-18
Update initial values for `prefix` and `suffix`

mfluehr:
cf. https://drafts.csswg.org/css-counter-styles-3/#counter-style-prefix


lahmatiy:
Your changes makes things better. However, I think it's a bad idea to add a comment to initial value. May be we can move comments to l10n?

mfluehr:
I see your point, @lahmatiy. Do you have a recommendation for precisely how that should work? Is it possible to have a l10n string as part of the initial value? Or maybe the comments can just be removed? I'm not sure they're actually all that necessary. (Also, I can't think of any other place where comments are used for initial values. Can you?)

lahmatiy:
@mfluehr My suggestion is to add to `l10n/css.json`:
```json
"\" \"": {
    "en-US": "\" \" (the empty string)"
},
"\". \"": {
    "en-US": "\". \" (full stop followed by a space)"
}
```
And left in `at-rules.json`:
```json
"initial": "\" \"",
...
"initial": "\". \"",.
```

wbamberg:
I've tested this in a local Kuma and the cssinfo macro seems happy with it, but I'm not sure why we are using `" "` here rather than `""`, which is what the spec has.

mfluehr:
I think you're right, @wbamberg, that the prefix should have no space.

PR #164, opened 2018-01-22
Update break-word

CShepartd:
Update break-word according to update https://developer.mozilla.org/en-US/docs/Web/CSS/word-break and specs https://drafts.csswg.org/css-text-3/#word-break-property


lahmatiy:
LGTM
Thank you for the links!

PR #167, opened 2018-01-23
adding overscroll-behavior and longhand properties

chrisdavidmills:
See https://wicg.github.io/overscroll-behavior/


lahmatiy:
LGTM

PR #169, opened 2018-01-31
attempting to add ::slotted pseudo-element and necessary syntaxes

chrisdavidmills:
I'm really not sure about the syntaxes.json additions, but through I've give it a go anyway. Always happy to learn.

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -111,7 +111,7 @@
     "syntax": "add | subtract | intersect | exclude"
   },
   "compound-selector": {
-    "syntax": "[ <type-selector>? <subclass-selector>*[ <pseudo-element-selector> <pseudo-class-selector>* ]* ]!"
+    "syntax": "[ <type-selector>? <subclass-selector>* [ <pseudo-element-selector> <pseudo-class-selector>* ]* ]!"
   },
   "compound-selector-list": {
     "syntax": "<compound-selector>#"
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -17,6 +17,15 @@
   "attr()": {
     "syntax": "attr( <attr-name> <type-or-unit>? [, <attr-fallback> ]? )"
   },
+  "attr-matcher": {
+    "syntax": "[ '~' | '|' | '^' | '$' | '*' ]? '='"
+  },
+  "attr-modifier": {
+    "syntax": "i"
+  },
+  "attribute-selector": {
+    "syntax": "'[' <wq-name> ']' | '[' <wq-name> <attr-matcher> [ <string-token> | <ident-token> ] <attr-modifier>? ']'"
+  },
   "auto-repeat": {
     "syntax": "repeat( [ auto-fill | auto-fit ] , [ <line-names>? <fixed-size> ]+ <line-names>? )"
   },
@@ -76,6 +85,9 @@
   },
   "circle()": {
     "syntax": "circle( [ <shape-radius> ]? [ at <position> ]? )"
+  },
+  "class-selector": {
+    "syntax": "'.' <ident-token>"
   },
   "clip-source": {
     "syntax": "<url>"
@@ -266,6 +278,9 @@
   "hue-rotate()": {
     "syntax": "hue-rotate( <angle> )"
   },
+  "id-selector": {
+    "syntax": "<hash-token>"
+  },
   "image": {
     "syntax": "<url> | <image()> | <image-set()> | <element()> | <cross-fade()> | <gradient>"
   },
@@ -392,6 +407,9 @@
   "namespace-prefix": {
     "syntax": "<ident>"
   },
+  "ns-prefix": {
+    "syntax": "[ <ident-token> | '*' ]? '|'"
+  },
   "number-percentage": {
     "syntax": "<number> | <percentage>"
   },
@@ -433,6 +451,12 @@
   },
   "position": {
     "syntax": "[ [ left | center | right ] || [ top | center | bottom ] | [ left | center | right | <length-percentage> ] [ top | center | bottom | <length-percentage> ]? | [ [ left | right ] <length-percentage> ] && [ [ top | bottom ] <length-percentage> ] ]"
+  },
+  "pseudo-class-selector": {
+    "syntax": "':' <ident-token> | ':' <function-token> <any-value> ')'"
+  },
+  "pseudo-element-selector": {
+    "syntax": "':' <pseudo-class-selector>"
   },
   "pseudo-page": {
     "syntax": ": [ left | right | first | blank ]"
@@ -557,6 +581,9 @@
   "step-timing-function": {
     "syntax": "step-start | step-end | steps(<integer>[, [ start | end ] ]?)"
   },
+  "subclass-selector": {
+    "syntax": "<id-selector> | <class-selector> | <attribute-selector> | <pseudo-class-selector>"
+  },
   "symbol": {
     "syntax": "<string> | <image> | <custom-ident>"
   },
@@ -611,10 +638,16 @@
   "type-or-unit": {
     "syntax": "string | integer | color | url | integer | number | length | angle | time | frequency | em | ex | px | rem | vw | vh | vmin | vmax | mm | q | cm | in | pt | pc | deg | grad | rad | ms | s | Hz | kHz | %"
   },
+  "type-selector": {
+    "syntax": "<wq-name> | <ns-prefix>? '*'"
+  },
   "var()": {
     "syntax": "var( <custom-property-name> [, <declaration-value> ]? )"
   },
   "viewport-length": {
     "syntax": "auto | <length-percentage>"
+  },
+  "wq-name": {
+    "syntax": "<ns-prefix>? <ident-token>"
   }
 }
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -98,6 +98,12 @@
   "compositing-operator": {
     "syntax": "add | subtract | intersect | exclude"
   },
+  "compound-selector": {
+    "syntax": "[ <type-selector>? <subclass-selector>*[ <pseudo-element-selector> <pseudo-class-selector>* ]* ]!"
+  },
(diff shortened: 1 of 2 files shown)


teoli2003:
Theoretically we should add 
type-selector
subclass-selector
pseudo-element-selector
and pseudo-class-selector

(and recursively so on).

But this case is a bit unusual as it goes into a lot of nested definitions. I'm not sure if we shouldn't consider type-selector, subclass-selector, pseudo-element-selector, pseudo-class-selector as basic types and give them a page.

@Elchi3  what do you think?

chrisdavidmills:
Hrm, I'm not really sure what to do here.

chrisdavidmills:
Oh, I see — we'd have to write pages and syntax entries for all the defs, subdefs, subsubdefs, etc.

Yeah, that sounds like a world of pain — we are talking about defs that go six or seven nested levels deep.

So I'd go for flattening the first 4 down as basic types ;-)

But still, I'm not 100% sure how I'd do that. This is a pretty complex job.

teoli2003:
We don't write pages for all the defs, subdefs, subsubdefs, only on the terminal (or flattened ones). If you have more than 2 pages here, don't do it. Likely not worth it.

teoli2003:
(I'm a bit swamped this week, so can't dig this more now)

wbamberg:
Let's try to get this merged.

`::slotted` is defined [in the spec](https://drafts.csswg.org/css-scoping-1/#selectordef-slotted) like this:

```
::slotted( <compound-selector-list> )
```

Iterating down into that would give us the following syntax entries to add:

```
<compound-selector-list> = <compound-selector>#
<compound-selector> = [ <type-selector>? <subclass-selector>* [ <pseudo-element-selector> <pseudo-class-selector>* ]* ]!
<type-selector> = <wq-name> | <ns-prefix>? '*'
<wq-name> = <ns-prefix>? <ident-token>
<ns-prefix> = [ <ident-token> | '*' ]? '|'
<subclass-selector> = <id-selector> | <class-selector> | <attribute-selector> | <pseudo-class-selector>
<id-selector> = <hash-token>
<class-selector> = '.' <ident-token>
<attribute-selector> = '[' <wq-name> ']' | '[' <wq-name> <attr-matcher> [ <string-token> | <ident-token> ] <attr-modifier>? ']'
<attr-matcher> = [ '~' | '|' | '^' | '$' | '*' ]? '='
<attr-modifier> = i
<pseudo-class-selector> = ':' <ident-token> | ':' <function-token> <any-value> ')'
<pseudo-element-selector> = ':' <pseudo-class-selector>
```

Adding these is easy. That leaves us with the following leaf entities:

```
<ident-token>
<hash-token>
<string-token>
<function-token>
<any-value>
```

I'm not sure whether these should be considered "basic types". [In our list of basic types](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Types) we include, for example, [`<string>`](https://developer.mozilla.org/en-US/docs/Web/CSS/string), which is described [in the specification](https://drafts.csswg.org/css-values-3/#strings) as "[corresponding] to the \<string-token\> production". So maybe we should interpret `<string-token>` as `<string>` and link to our docs page for that?

Similarly, the CSS Values spec lists an [`<ident>`](https://drafts.csswg.org/css-values-3/#typedef-ident) type described as "conforming to the \<ident-token\> grammar", although we don't have an `<ident>` page on MDN.

However I didn't find anything similar for the other tokens.

But, stepping back: I do wonder how much value to users of MDN all this stuff would add? What if we just define the syntax like:

```
::slotted( selector )
```

and leave it at that?

There is somewhat precedent for doing this. For example, look at the next-sibling combinator. The [spec for this](https://drafts.csswg.org/selectors-4/#adjacent-sibling-combinators) says: "The next-sibling combinator is made of the “plus sign” (U+002B, +) code point that separates two compound selectors." But for this item we don't include `<compound-selector>` in the syntax, we just describe it like this:

```
"syntax": "A + B",
```
(https://github.com/mdn/data/blob/master/css/selectors.json#L43)

Is there any reason not to do something analogous here?

@chrisdavidmills , what do you think?

wbamberg:
OK, so I didn't see this: https://github.com/mdn/data/pull/186#discussion_r172397050 or this: https://github.com/mdn/data/issues/190 which cover basically the same ground.

Then perhaps what we should do for now is add these syntaxes:

```
<compound-selector-list> = <compound-selector>#
<compound-selector> = [ <type-selector>? <subclass-selector>* [ <pseudo-element-selector> <pseudo-class-selector>* ]* ]!
<type-selector> = <wq-name> | <ns-prefix>? '*'
<wq-name> = <ns-prefix>? <ident-token>
<ns-prefix> = [ <ident-token> | '*' ]? '|'
<subclass-selector> = <id-selector> | <class-selector> | <attribute-selector> | <pseudo-class-selector>
<id-selector> = <hash-token>
<class-selector> = '.' <ident-token>
<attribute-selector> = '[' <wq-name> ']' | '[' <wq-name> <attr-matcher> [ <string-token> | <ident-token> ] <attr-modifier>? ']'
<attr-matcher> = [ '~' | '|' | '^' | '$' | '*' ]? '='
<attr-modifier> = i
<pseudo-class-selector> = ':' <ident-token> | ':' <function-token> <any-value> ')'
<pseudo-element-selector> = ':' <pseudo-class-selector>
```

...and just omit the definition of the leaf items.

chrisdavidmills:
OK, that sounds like a fair solution. I have made the changes you requested.

lahmatiy:
On low level CSS have a token-based grammar. Tokens are producing a types, that produces other types and so on. The list of tokens can be found in [Tokenization section](https://drafts.csswg.org/css-syntax/#tokenization) of [CSS Syntax Module Level 3](https://drafts.csswg.org/css-syntax/).
In some cases tokens are equal to a generic types. For example, CSS Values and Units L3 claims:
> Strings are denoted by `<string>` and consist of a sequence of characters delimited by double quotes or single quotes. They correspond to the `<string-token>` production in the CSS Syntax Module [CSS3SYN].

and

> Identifiers, generically denoted by `<ident>`, consist of a sequence of characters conforming to the `<ident-token>` grammar.

However, that's not the common case. For example, `<function-token>` can be described as a pair of `<ident-token>` and `<(-token>` with no whitespaces between the tokens. `<hash-token>` is more complicated, since it represents `<hex-color>` and things like `<id-selector>`, the difference in a value of additional flag (which influences on parsing rules and other things):

> ... hash tokens have a type flag set to either "id" or "unrestricted". The type flag defaults to "unrestricted" if not otherwise set.

Much complicated things like `<any-value>`. Spec claims:

> The `<declaration-value>` production matches any sequence of one or more tokens, so long as the sequence does not contain `<bad-string-token>`, `<bad-url-token>`, unmatched `<)-token>`, `<]-token>`, or `<}-token>`, or top-level `<semicolon-token>` tokens or `<delim-token>` tokens with a value of "!". It represents the entirety of what a valid declaration can have as its value.
> The `<any-value>` production is identical to `<declaration-value>`, but also allows top-level `<semicolon-token>` tokens and `<delim-token>` tokens with a value of "!". It represents the entirety of what valid CSS can be in any context.

Such things can't be represented with a regular grammars, that's a set of rules.

Few notes on @wbamberg comments:

> I do wonder how much value to users of MDN all this stuff would add? What if we just define the syntax like ...

It depends on what purposes of dictionaries. If main goal is provide a documentation for a users who learn CSS, then simplification is a good thing. However, those syntaxes can be used (and using already) for various purpose, where such simplification is a bad thing. According to [Selectors Level 4](https://www.w3.org/TR/selectors-4/#structure):

> ... The term selector can refer to a simple selector, compound selector, complex selector, or selector list. The subject of a selector is any element that selector is defined to be about; that is, any element matching that selector.

So, if we set `::slotted( selector )` as a syntax, it's quite unclear what type of "selector" used in that context. It's also useless for a tools like CSSTree's syntax validator. I believe syntaxes should reflect spec definitions first. If a user friendly description is needed, an additional field can be added for that purpose.

> For example, look at the next-sibling combinator. The spec for this says: "The next-sibling combinator is made of the “plus sign” (U+002B, +) code point that separates two compound selectors." But for this item we don't include <compound-selector> in the syntax, we just describe it like this:
> "syntax": "A + B",

Well, `next-sibling combinator` is a *combinator*. It's a part of a complex selector, but not an any type of selector. So placing it into `selector.json` is a bit strange, that's not the case.

I think we will add all the syntaxes from Selector L4 soon or later.

wbamberg:
@lahmatiy , sorry, yes, I was just speaking from the point of view of MDN pages. particularly wrt the suggestion that we should write MDN pages for low-level items in the syntax definition, just so we have something to point to from the lower reaches of a highly abstract formal syntax definition. I don't think that would be a helpful thing to do for users of MDN.

Of course the situation is different for you. But in that case we need to understand your requirements here. Does the proposal outlined here (https://github.com/mdn/data/pull/169#issuecomment-388500845) and currently offered by Chris in the PR work for you at the moment, or do you think we need to find a place to document the syntax of the tokens in the data?

css/syntaxes.json
```diff
@@ -98,6 +110,12 @@
   "compositing-operator": {
     "syntax": "add | subtract | intersect | exclude"
   },
+  "compound-selector": {
+    "syntax": "[ <type-selector>? <subclass-selector>*[ <pseudo-element-selector> <pseudo-class-selector>* ]* ]!"
```
wbamberg: There should be a space here, so not:
```
<subclass-selector>*[
```
...but
```
<subclass-selector>* [
```

I don't know if whitespace changes the meaning, but we might as well fix this.

lahmatiy:
@wbamberg LGTM.
`<*-token>` syntaxes are not a subject of those dictionaries since they are "atoms" of CSS syntax. The same for `<any-value>` and some others.
Probably we need mark some syntaxes with a new flag in future, since there is the additional restriction "White space is forbidden between tokens except..." – it's unusual for a syntaxes.
You could merge this PR since it reflects a spec. If I found an issue I'll back with a PR ;)

PR #176, opened 2018-02-09
Mdn porting

jameshkramer:
Creating pull request per @lahmatiy.

PR #177, opened 2018-02-13
Fixes per review by @lahmatiy

jameshkramer:


jameshkramer:
@lahmatiy, @Elchi3, and @erikadoyle: the Microsoft work is now in #177. I just pushed the changes called for in #176.

css/properties.json
```diff
@@ -173,22 +173,22 @@
...
-    "initial": "false",
+    "initial": "none",
     "appliesto": "nonReplacedBlockAndInlineBlockElements",
     "computed": "asSpecified",
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
   "-ms-filter": {
-    "syntax": "<ms-filter-specification>",
+    "syntax": "<string>+",
```
lahmatiy: `<string>+` -> `<string>`
According to [MSDN](https://msdn.microsoft.com/en-us/library/ms530752(v=vs.85).aspx), that's a single string with comma separated values inside the string. No multiplier is needed.

l10n/css.json
```diff
@@ -1474,12 +1474,12 @@
...
   "referToLineBox": {
-    "en-US": "Calculated with respect to the line box"
+    "en-US": "Calculated with respect to the width of the line box"
   },
   "lengthsAsPercentages": {
     "en-US": "The lengths may be specified as percentages"
   },
   "exclusionElements": {
-    "en-US": "exclusion elements"
+    "en-US": "exclusion elements (created by -ms-wrap-flow)"
```
lahmatiy: Not sure `-ms-wrap-flow` should be mentioned

lahmatiy:
Thank you for the fixes!
Please, merge `master` into you branch (e.g. some `display-*` properties were remove and `overscroll-behavior` was added).
A few comments after a cursory review:
- `-ms-content-zoom-limit-min`/`-ms-content-zoom-limit-max` percentages values are still min/maxZoomFactor – but should be `magnificationScale` or something like that.
- I realised that `zoom` property is absent in the dictionary. It would be nice to add it too ([MDN page](https://developer.mozilla.org/en-US/docs/Web/CSS/zoom)).
- `-ms-scrollbar-arrow-color`, `-ms-scrollbar-base-color`, `-ms-scrollbar-face-color`, `-ms-scrollbar-highlight-color`, `-ms-scrollbar-track-color` are missed. And we need to find `-ms-
scrollbar-*-color` initial values. I suppose some non-standard named colors are used for those properties.

jameshkramer:
@lahmatiy, thank you very much for the review. Here are my comments questions:

- I changed the syntax for ms-filter to <string>, as you indicated (not yet pushed).

- I added zoom to properties.json (assuming that is what you meant by the dictionary). I've modified the page to include {{cssinfo}} and {{csssyntax}}. I'll add a link to our extensions page for zoom.

- I removed the mention of -ms-wrap-flow from exclusionElements. I had only put it there in response to one of your earlier comments, which evidently I misinterpreted.

- Properties -ms-scrollbar-arrow-color, -ms-scrollbar-base-color, -ms-scrollbar-face-color, -ms-scrollbar-highlight-color, and -ms-scrollbar-track-color are not on my list to port to MDN. Perhaps @erikadoyle will comment on this. Regardless, I would like this issue to be separated away from the current effort to change the DB; i.e. I am hoping to get approval on the current work even if it does not include these properties.

- We not know the initial values for ms-scrollbar-3dlight, ms-scrollbar-darkshadow-color, and ms-scrollbar-shadow-color. There is no specification, and the PDL says "undefined". Of course we believe that IE implements initial values in its various versions, but my attempts to discover them experimentally did not succeed. Our main goal is to port existing MSDN content, so we do not want to spend much time at all fixing content; thus we want to stop our efforts to determine these values. We do welcome your suggestions about how best document the situation.

- I am confused about the maxZoomFactor/minZoomFactor issue. The terms "zoom" and "zoom factor" seem to  be well-established, and I see them used in the min-zoom, max-zoom, and zoom pages, as well as in the HTML specification. So I don't understand the need to change maxZoomFactor and minZoomFactor. I am not asking for an explanation (though I would appreciate it), but I do want to be sure that you want a change. Also, do you want to change the text corresponding to the names, or just the names themselves?

- The invalid changes to the master file were due to my inexperience. I see what happened and will solve it.

Thanks,
Jim

jameshkramer:
@lahmatiy, I talked to @erikadoyle about the other 5 properties. Their omission is an oversight. We'll do them, perhaps in this PR.

lahmatiy:
Thank you for comments. My thoughts:

> Properties -ms-scrollbar-arrow-color, -ms-scrollbar-base-color, -ms-scrollbar-face-color, -ms-scrollbar-highlight-color, and -ms-scrollbar-track-color are not on my list to port to MDN.

Those properties are a part of `-ms-scrollbar-*` family. It's a quite odd to add 3 of 8 properties of the family. The more they are similar, the difference is only in the name.

> We not know the initial values for ms-scrollbar-3dlight, ms-scrollbar-darkshadow-color, and ms-scrollbar-shadow-color. ... We do welcome your suggestions about how best document the situation.

Yep, that's a complicated case. I tried to retrieve those values, and did not succeed too. I suppose a IE uses system colors. Probably the truth can be found in source codes. Unfortunately, they aren't open sourced, and I believe even with sources it can be impossible. As a solution, we can use `dependsOnUserAgent` as a value (although it depends on OS, but I'm not sure).

> Our main goal is to port existing MSDN content, so we do not want to spend much time at all fixing content

I understand, but we should try ;) Broken or incomplete data is not useful.

> I am confused about the maxZoomFactor/minZoomFactor issue. The terms "zoom" and "zoom factor" seem to be well-established, and I see them used in the min-zoom, max-zoom, and zoom pages, as well as in the HTML specification. So I don't understand the need to change maxZoomFactor and minZoomFactor. I am not asking for an explanation (though I would appreciate it), but I do want to be sure that you want a change. Also, do you want to change the text corresponding to the names, or just the names themselves?

`percentage` value should explain what a percentage value refers to, from which the percentage is calculated. That's may be confused, because those properties are limit the value of another one. Take a look at `min-width`/`max-width` or similar, they describe a percentage value as `referToWidthOfContainingBlock` not as `minWidth`/`maxWidth`.

Waiting for your further commits.

jameshkramer:
@lahmatiy, I talked to @erikadoyle about the other 5 properties. Their omission is an oversight. We'll do them, probably in this PR.

jameshkramer:
Created a new branch. There will be a new PR.

PR #180, opened 2018-02-22
updating syntax for align-content, justify-content, and place-content

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1430817.

There is one aspect that I'm not happy with as yet. The appliesto values for these properties should be 'multicol containers, flex containers, and grid containers' for justify-content and place-content, and 'block containers, multicol containers, flex containers, and grid containers' for align-content — how do I represent that in the properties JSON?

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -1428,7 +1428,7 @@
     "status": "nonstandard"
   },
   "align-content": {
-    "syntax": "flex-start | flex-end | center | space-between | space-around | space-evenly | stretch",
+    "syntax": "normal | <baseline-position> | <content-distribution> | <overflow-position> ? <content-position>",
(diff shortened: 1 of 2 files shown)


css/properties.json
```diff
@@ -6121,6 +6121,21 @@
     "order": "oneOrTwoValuesLengthAbsoluteKeywordsPercentages",
     "status": "standard"
   },
+  "place-content": {
+    "syntax": "<'align-content'> <'justify-content'>?",
```
emilio: Just note that this has the gotcha that a single `<align-content>` value that wouldn't be valid as `<justify-content>` is invalid.

PR #182, opened 2018-02-26
updating the syntax of align-self and align-items as per latest box a…

chrisdavidmills:
…lignment spec.

See emilio's comment at https://bugzilla.mozilla.org/show_bug.cgi?id=1430817#c43

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -1467,7 +1467,7 @@
       "CSS Flexible Box Layout"
     ],
     "initial": "auto",
-    "appliesto": "flexItemsAndInFlowPseudos",
+    "appliesto": "flexItemsGridItemsAndAbsolutelyPositionedBoxes",
     "computed": "autoOnAbsolutelyPositionedElementsValueOfAlignItemsOnParent",
     "order": "uniqueOrder",
     "status": "standard"
--- a/css/properties.json
+++ b/css/properties.json
@@ -1451,14 +1451,14 @@
     "groups": [
       "CSS Flexible Box Layout"
     ],
-    "initial": "stretch",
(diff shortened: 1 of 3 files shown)


lahmatiy:
Spec references:
- https://www.w3.org/TR/css-align-3/#propdef-align-self
- https://www.w3.org/TR/css-align-3/#align-items-property

Currently `align-self` has `flexItemsAndInFlowPseudos` as a value for `applyto`. Per spec it should be `flex items, grid items, and absolutely-positioned boxes`. Can we fix it?

chrisdavidmills:
I'd be happy to fix this, if I can work out how.

I can see that the "appliesto" choices are defined here:
https://github.com/mdn/data/blob/master/css/properties.schema.json#L153

But surely I can't just add "flexItemsgridItemsAndAbsolutelyPositionedBoxes" here and in the property definition, and just expect it to work? Don't I need to define an actual string that this value maps to somewhere, to display on the actual page? I couldn't find where that might be stored.

including @Elchi3 on this tread, as he probably knows ;-)

lahmatiy:
> I can't just add "flexItemsgridItemsAndAbsolutelyPositionedBoxes"

@chrisdavidmills Don't worry, you can ;) You can add such value to `applyto`, to schema and to `l10n/css.json`. It should works then.

chrisdavidmills:
OK, I've tried fixing this up - thanks for your help @lahmatiy !

lahmatiy:
@chrisdavidmills Good work! And welcome :)

pkuczynski:
@wbamberg is this ready to merge?

PR #186, opened 2018-02-28
adding matches pseudo-class

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1440868

### DIFF
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -272,7 +272,7 @@
     ],
     "status": "standard"
   },
-  ":matches()": {
+  ":matches": {
     "syntax": ":matches( <selector-list> )",
     "groups": [
       "Pseudo-classes",
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -273,7 +273,7 @@
     "status": "standard"
   },
   ":matches()": {
-    "syntax": ":matches( selector[, selector]* )",
+    "syntax": ":matches( <selector-list> )",
     "groups": [
       "Pseudo-classes",
       "Selectors"
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -272,6 +272,14 @@
     ],
     "status": "standard"
   },
+  ":matches": {
+    "syntax": ":matches( selector[, selector]* )",
+    "groups": [
+      "Pseudo-classes",
+      "Selectors"
+    ],
+    "status": "standard"
+  },
   ":not": {
     "syntax": ":not( <selector># )",
     "groups": [
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -272,7 +272,7 @@
     ],
     "status": "standard"
   },
-  ":matches": {
+  ":matches()": {
     "syntax": ":matches( selector[, selector]* )",
     "groups": [
       "Pseudo-classes",
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -12,7 +12,7 @@
     "syntax": "scroll-position | contents | <custom-ident>"
   },
   "any-value": {
-    "[!]+"
+    "syntax": "[!]+"
   },
   "attachment": {
     "syntax": "scroll | fixed | local"
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -11,11 +11,23 @@
   "animateable-feature": {
     "syntax": "scroll-position | contents | <custom-ident>"
   },
+  "any-value": {
+    "[!]+"
+  },
   "attachment": {
     "syntax": "scroll | fixed | local"
   },
   "attr()": {
     "syntax": "attr( <attr-name> <type-or-unit>? [, <attr-fallback> ]? )"
+  },
+  "attr-matcher": {
+    "syntax": "[ '~' | '|' | '^' | '$' | '*' ]? '='"
+  },
+  "attr-modifier": {
+    "syntax": "i"
+  },
+  "attribute-selector": {
+    "syntax": "'[' <wq-name> ']' | '[' <wq-name> <attr-matcher> [ <string-token> | <ident-token> ] <attr-modifier>? ']'"
   },
   "auto-repeat": {
     "syntax": "repeat( [ auto-fill | auto-fit ] , [ <line-names>? <fixed-size> ]+ <line-names>? )"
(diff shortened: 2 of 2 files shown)


css/selectors.json
```diff
@@ -272,6 +272,14 @@
     ],
     "status": "standard"
   },
+  ":matches()": {
+    "syntax": ":matches( selector[, selector]* )",
```
lahmatiy: `selector[, selector]*` -> `<selector>#`
Actually it takes a `<selector-list>` (https://www.w3.org/TR/selectors-4/#matches), but this syntax is not defined yet. Grammar can be taken here: https://www.w3.org/TR/selectors-4/#grammar

chrisdavidmills:
@lahmatiy So, I tried adding the necessary syntax definitions. That was a serious rabbit hole to fall down ;-)

I took all the syntax definitions straight from the spec, except for:

* <ident-token>
* <hash-token>
* <function-token>
* <string-token>
* <any-value>

These are just described and not actually shown, so I'm not 100% sure if I've got these right.

css/selectors.json
```diff
@@ -272,6 +272,14 @@
     ],
     "status": "standard"
   },
+  ":matches()": {
```
ExE-Boss: This should be [`:matches`](https://developer.mozilla.org/en-US/docs/Web/CSS/:matches).

css/syntaxes.json
```diff
@@ -11,12 +11,24 @@
   "animateable-feature": {
     "syntax": "scroll-position | contents | <custom-ident>"
   },
+  "any-value": {
+    "[!]+"
```
lahmatiy: I don't know what means such syntax. Could you explain?

css/syntaxes.json
```diff
@@ -233,6 +260,9 @@
   "frequency-percentage": {
     "syntax": "<frequency> | <percentage>"
   },
+  "function-token": {
```
lahmatiy: `function-token` is a token type and out of scope of this dict.
Btw, it has another syntax

css/syntaxes.json
```diff
@@ -254,6 +284,9 @@
   "grid-line": {
     "syntax": "auto | <custom-ident> | [ <integer> && <custom-ident>? ] | [ span && [ <integer> || <custom-ident> ] ]"
   },
+  "hash-token": {
```
lahmatiy: `hash-token` is a token type and out of scope of this dict.
Btw, it has another syntax

css/syntaxes.json
```diff
@@ -269,6 +302,12 @@
   "hue-rotate()": {
     "syntax": "hue-rotate( <angle> )"
   },
+  "id-selector": {
+    "syntax": "<hash-token>"
+  },
+  "ident-token": {
```
lahmatiy: `ident-token` is a token type and out of scope of this dict.
Btw, it has another syntax

css/syntaxes.json
```diff
@@ -560,6 +599,9 @@
   "nth": {
     "syntax": "<an-plus-b> | even | odd"
   },
+  "ns-prefix": {
+    "syntax": "[ <ident-token> | '*' ]? '|'"
+  }
```
lahmatiy: Comma is missed

css/syntaxes.json
```diff
@@ -716,6 +767,12 @@
   "step-timing-function": {
     "syntax": "step-start | step-end | steps(<integer>[, [ start | end ] ]?)"
   },
+  "string-token": {
```
lahmatiy: `string-token` is a token type and out of scope of this dict.
Btw, it has another syntax

css/syntaxes.json
```diff
@@ -269,6 +302,12 @@
   "hue-rotate()": {
     "syntax": "hue-rotate( <angle> )"
   },
+  "id-selector": {
+    "syntax": "<hash-token>"
+  },
+  "ident-token": {
+    "syntax": "[0-10FFFF]*"
+  }
```
lahmatiy: Comma is missed

css/syntaxes.json
```diff
@@ -11,12 +11,24 @@
   "animateable-feature": {
     "syntax": "scroll-position | contents | <custom-ident>"
   },
+  "any-value": {
+    "syntax": "[!]+"
```
lahmatiy: I don't know what means such syntax. Could you explain?

css/syntaxes.json
```diff
@@ -11,12 +11,24 @@
   "animateable-feature": {
     "syntax": "scroll-position | contents | <custom-ident>"
   },
+  "any-value": {
```
lahmatiy: I believe it's a generic that out of scope of this dict

css/syntaxes.json
chrisdavidmills: So I should remove these then...? How should this be represented then in our data? Some kind of link to the spec definition of token types?

The syntax definition I included was basically a guess based on my knowledge of related syntax, but I didn't really know what I was doing.

Same comment for the other token types.

css/syntaxes.json
chrisdavidmills: Heh, I was guessing what it should possibly be, based on reading this defintion https://drafts.csswg.org/css-syntax-3/#typedef-any-value

css/syntaxes.json
lahmatiy: Those tokens are defined in [CSS Syntax Module Level 3](https://drafts.csswg.org/css-syntax/) and shouldn't be presented in `syntax.json`. Because they are something like a generic, as `<string>` or `<number>` but a bit low-level.

css/syntaxes.json
lahmatiy: Oh, well, that' diffidently incorrect. `[!]+` means one or more exclamation marks and nothing else ;)
`<any-value>` can't be represented with such syntax, since it means any valid CSS value. Actually it's a generic and should be excluded from `syntax.json`.

chrisdavidmills: Thank you again ;-)

I have opened up a separate issue to track thisL: https://github.com/mdn/data/issues/190

chrisdavidmills:
@teoli2003 can I get your advice on how we can make progress on this one?

pkuczynski:
@lahmatiy if we remove all you listed above, would it be ready to go?

ExE-Boss:
This needs rebasing

ExE-Boss:
I got fed up with waiting and opened #284.

chrisdavidmills:
I'm really glad you did — you've actually come up with a really simple solution to this, whereas my attempt was horribly overcomplicated and we got stalled on it.

PR #188, opened 2018-02-28
Sort css.json alphabetically

pkuczynski:

PR #185, opened 2018-02-28
Add justify-items CSS property

pkuczynski:
Based on:

https://drafts.csswg.org/css-align-3/#propdef-justify-items
https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content

Fixes #149, https://github.com/csstree/csstree/issues/81

PR #189, opened 2018-03-01
Add gap, row-gap and update column-gap CSS properties

pkuczynski:
Based on:
* https://www.w3.org/TR/css-align-3/#gap-shorthand
* https://www.w3.org/TR/css-align-3/#column-row-gap
* https://www.w3.org/TR/css-align-3/#gap-legacy

Additionally marking `grid-gap`, `grid-column-gap`, `grid-row-gap` as obsolete based on https://www.w3.org/TR/2016/CR-css-grid-1-20160929/

Refs https://github.com/csstree/csstree/issues/81


css/properties.json
```diff
@@ -4152,6 +4153,30 @@
...
+    "inherited": false,
+    "animationType": [
+      "row-gap",
+      "column-gap"
+    ],
+    "percentages": "no",
+    "groups": [
+      "CSS Box Alignment"
+    ],
+    "initial": [
```
lahmatiy: Should be 
```
"initial": [
      "row-gap",
      "column-gap"
]
```

css/properties.json
```diff
@@ -3332,17 +3332,18 @@
...
+    "syntax": "normal | <length-percentage>",
     "media": "visual",
     "inherited": false,
     "animationType": "lpc",
-    "percentages": "referToContentWidthOfMulticolElement",
+    "percentages": "referToDimensionOfContentArea",
     "groups": [
-      "CSS Columns"
+      "CSS Columns",
+      "CSS Box Alignment"
```
lahmatiy: This group is not defined yet. I guess that's will be fixed when #185 will be merged.

pkuczynski: That's correct! Btw. how long it usually takes to get PR merged in?

css/properties.json
pkuczynski: I was thinking about it, but documentation on w3c says something different. Please check...

lahmatiy: Looks like a mistake in the spec. A shorthand should have the same initial values as its corresponding properties.
We need to fill out an issue on the problem in the spec's tracker.

pkuczynski:
Conflicts solved but #188 needs to be merged first

css/properties.json
pkuczynski: Fixed on my side. Not sure how to fill out an issue on the problem in the spec's tracker.

css/properties.json
pkuczynski: Yes, and it's already merged.

css/properties.json
lahmatiy: You can find a link to issue tracker in the beginning of spec: https://github.com/w3c/csswg-drafts/labels/css-align-3
It's hosted on GitHub, just create an issue with prefix `[css-align] ` and describe a problem: `gap` is a shorthand for `row-gap` and `column-gap`, but its initial value differ with initial values of those properties.

pkuczynski: Done https://github.com/w3c/csswg-drafts/issues/2504

pkuczynski: Should we merge this PR in?

lahmatiy: Here is an answer on your issue:
> You are right, it should say "see individual properties". This has already been corrected in the latest draft: https://drafts.csswg.org/css-align/#gap-shorthand

I was right 😎So we can merge it without any hesitation

pkuczynski: Yeay! First we need to merge #188 - can u have a look at it?

pkuczynski:
This is ready to be merged! @wbamberg?

l10n/css.json
```diff
@@ -384,9 +384,8 @@
     "fr": "telle que spécifiée, sauf avec &lt;resolution&gt; éventuellement modifiée lors du calcul par la valeur 'snap'",
     "ru": "как указано, за исключением &lt;resolution&gt; может изменить значением 'snap'"
   },
-  "asSpecifiedWithLengthsMadeAbsolute": {
-    "de": "wie angegeben, wobei Längen in absolute Werte umgewandelt werden",
-    "en-US": "as specified with lengths made absolute"
+  "asSpecifiedWithLengthsAbsoluteAndNormalComputingToZeroExceptMultiColumn": {
+    "en-US": "as specified, with <length>s made absolute, and normal computing to zero except on multi-column elements"
```
wbamberg: I think this should encode `<` and `>` in `<length>`, shouldn't it?

css/properties.json
pkuczynski: #188 merged so this can go in too!

pkuczynski:
@wbamberg would this PR have its lucky day? ;)

css/properties.json
```diff
@@ -4247,7 +4272,9 @@
       "grid-auto-columns",
       "grid-auto-flow",
       "grid-column-gap",
-      "grid-row-gap"
+      "grid-row-gap",
+      "column-gap",
+      "row-gap"
```
wbamberg: I'm still trying to learn this, so forgive me, but... AIUI this change is to reflect the fact that the spec now uses `row-gap` and `column-gap` instead of `grid-row-gap` and `grid-column-gap`, presumably to reflect the fact that these properties are now in the box alignment spec and are thus available for flex items and multicol. Is that right?

-> https://drafts.csswg.org/css-grid/#change-2016-grid-gap

In this case should we also remove `grid-row-gap` and `grid-column-gap` from the data?

css/properties.json
```diff
@@ -4258,7 +4285,9 @@
...
```
wbamberg: Same question as above.

l10n/css.json
pkuczynski: You mean like `&gt;` type of thing?

css/properties.json
pkuczynski: I actually only added this two values and did not remove anything, but you right, they should be gone. However, I assume some browsers would support both. So what should we do?

I just pushed a change marking `grid-gap`, `grid-column-gap`, `grid-row-gap` as obsolete based on https://www.w3.org/TR/2016/CR-css-grid-1-20160929/

css/properties.json
pkuczynski: Same answer ;)

l10n/css.json
wbamberg: Yes, exactly. We do this in other places, e.g. https://github.com/mdn/data/blob/master/l10n/css.json#L163, and we also include markup in these strings, e.g. https://github.com/mdn/data/blob/master/l10n/css.json#L176, so it seems like we're just inserting these strings into the HTML output for MDN.

PR #195, opened 2018-04-05
Add ::slotted selector.

jpmedley:


lahmatiy:
We have got another PR about `::slotted`, see #169

PR #193, opened 2018-04-05
 Normalize media field values in css/properties.json and css/at-rules.json

pkuczynski:
Fixes #63

PR #196, opened 2018-04-06
 Generally improved validation and added l10n validation

pkuczynski:
* add validation for `l10n/css.json`
* verify unused keys in `l10n/css.json` - found a couple of unused keys and deleted them - was I right about it?
* move all definitions into common `definitions.js`
* generally improve the schema validation in few places

It would be possible to implement similar technique done for `l10n` also to `definitions#enums`, so there would be no need to manually define those enums, but instead validate if values exist in `l10n` and vice versa. What do you think about this idea?


frenic:
Maybe sort order should be linted as well?

pkuczynski:
@frenic can you merge this PR?

frenic:
I wish! 😆 I don't have the privilege to do that.

wbamberg:
I do, but this is a big (though impressive!) change, and I probably won't have time to review it today.

@frenic, you can definitely help me by reviewing it, if you have the time.

BTW @pkuczynski , @lahmatiy , I want to make a new release of this repo as soon as possible, are there any open PRS you would really like to have merged before I do so? (or conversely, any potentially risky ones that you would *not* like merged?)

test/lint/checkStyle.js
```diff
@@ -0,0 +1,19 @@
...
+function checkStyle (filename) {
+  var noErrors = true
+  var actual = fs.readFileSync(filename, 'utf-8').trim()
+  var expected = JSON.stringify(JSON.parse(actual), null, 2)
+
+  if (actual === expected) {
+    console.log('  Style – OK')
+  } else {
+    noErrors = false
+    console.log('  Style – Error on line ' + jsonDiff(actual, expected))
```
frenic: It would be nice maybe if this displayed all the diffs rather than just one line at the time.

frenic:
#189 would be nice to have

frenic:
> This PR doesn’t really change data, except l10n where I removed unused strings.

I meant that I haven't worked with JSON data to this extent before. So my knowledge about schemas is a bit limited.

frenic:
> There is a good question if we should not remove the whole enums in the definition.json and implement it differently using the same technique as for serching for unused translations? That would be my choice...

That sounds like a major change? @wbamberg In that case maybe it should be in another PR so it can be discussed separately?

pkuczynski:
@wbamberg #193 and #189 are good candidates to merge before the release I think.

wbamberg:
Please note that https://github.com/mdn/data/pull/189 needs a[ fix in l10n/css.json before](https://github.com/mdn/data/pull/189/files/87c80547b934fac1b37b060b243a2c655128474d#diff-8cf24efe85c7978d105a5046315efc4b) I can merge it. I think the other issue there (removing grid-row-gap and grid-column-gap) doesn't have to block merging though.

css/definitions.json
```diff
@@ -1,69 +1,366 @@
...
+  },
+  "propertyList": {
+    "type": "array",
+    "minItems": 1,
+    "uniqueItems": true,
+    "items": {
+      "$ref": "properties#/names"
+    }
+  },
+  "enums": {
```
lahmatiy: Why we need to split types for paths `enums/names/*` and `enums/text/*`? I believe it can be a plain list of types, i.e. `groups` instead of `enums/names/groups` and `initial` instead of `enums/text/initial`

test/lint.js
```diff
@@ -1,92 +1,26 @@
-var fs = require('fs');
-var path = require('path');
-var Ajv = require('ajv');
-var ajv = new Ajv({ $data: true, allErrors: true });
-var dictPaths = ['api', 'css', 'l10n'];
-var hasErrors = false;
+var fs = require('fs')
```
lahmatiy: It's not fair, you have changed a code style 😢

test/lint/checkStyle.js
lahmatiy: It's ok, `jsonDiff` returns all the diffs.

css/at-rules.json
```diff
@@ -54,7 +54,7 @@
       "prefix": {
         "syntax": "<symbol>",
         "media": "all",
-        "initial": "\"\"",
```
lahmatiy: I believe it should be `"\"\""` since it's a valid CSS value

css/at-rules.json
```diff
@@ -81,7 +81,7 @@
       "suffix": {
         "syntax": "<symbol>",
         "media": "all",
-        "initial": "\". \"",
```
lahmatiy: I believe it should be `"\". \""` since it's a valid CSS value

css/properties.json
```diff
@@ -188,7 +188,7 @@
     "groups": [
       "Microsoft Extensions"
     ],
-    "initial": "\"\"",
```
lahmatiy: I believe it should be `"\"\""` since it's a valid CSS value

css/definitions.json
pkuczynski: The reason why I split them, was because I am cross-checking `enums/text` with `l10n/css`. So if the key is missing in l10n or l10n has keys which does not exists in this list, the build will fail. This way I found some unused l10n strings. I felt like it is a good idea to keep things tidy.

I was also considering on dumping the whole `enum/text` and simply pull the values in js directly from individual props in `*.json`. This would limit the overhead of adding the new values into `definitions.js`. What do you think?

test/lint.js
pkuczynski: Apologies. That was my editor default settings. Do you want me to bring back the semicolons?

css/at-rules.json
pkuczynski: I changed this so it will place nicely with l10n check. Can bring it back if you want?

wbamberg:
@pkuczynski , I'm sorry to have not reviewed this PR. It seems from your description that the PR does a few different things, would it be possible to split it up into separate PRs that each do one thing? That would greatly help me find the time to digest and review it.

pkuczynski:
Shame it never got merged after 4y

PR #198, opened 2018-04-09
Add some missing items to selectors.json.

jpmedley:
Makes sure all the items listed in [pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) have an entry in selectors.json.

### DIFF
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -192,7 +192,7 @@
     "status": "standard"
   },
   ":focus-visible": {
-    "syntax": ":focus",
+    "syntax": "focus-visible",
     "groups": [
       "Pseudo-classes",
       "Selectors"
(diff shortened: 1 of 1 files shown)


css/selectors.json
```diff
@@ -191,6 +191,14 @@
     ],
     "status": "standard"
   },
+  ":focus-visible": {
+    "syntax": ":focus",
```
wbamberg: Should this be `:focus-visible`?

css/selectors.json
```diff
@@ -199,6 +207,22 @@
     ],
     "status": "experimental"
   },
+  ":host": {
+    "syntax": ":hover",
```
wbamberg: Should this be `:host`?

css/selectors.json
```diff
@@ -199,6 +207,22 @@
...
+    "groups": [
+      "Pseudo-classes",
+      "Selectors"
+    ],
+    "status": "experimental"
+  },
+  ":host-context": {
+    "syntax": ":hover",
```
wbamberg: Should this be `:host-context`?

jpmedley:
PTAL

PR #201, opened 2018-04-10
Add `text` and `border` keywords to background-clip

frenic:
https://drafts.csswg.org/css-backgrounds-4/#background-clip

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -48,7 +48,7 @@
     "syntax": "normal | multiply | screen | overlay | darken | lighten | color-dodge | color-burn | hard-light | soft-light | difference | exclusion | hue | saturation | color | luminosity"
   },
   "box": {
-    "syntax": "border-box | padding-box | content-box"
+    "syntax": "border-box | padding-box | content-box | text | border"
   },
   "br-style": {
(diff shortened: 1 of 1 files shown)


wbamberg:
@teoli2003 , @chrisdavidmills , what's our policy for when to add new syntax? The linked spec has a big red box saying:

> Not Ready For Implementation
>
> 
> This spec is not yet ready for implementation. It exists in this repository to record the ideas and promote discussion.
> 
> Before attempting to implement this spec, please contact the CSSWG at www-style@w3.org.
> 

At what point in the standards process should we add new features to MDN (and mdn/data)?

wbamberg:
Then I think in this case we should not add the syntax, and should close this PR. Because the spec explicitly tells us that it is not ready for implementation (i.e. not just that noone _has_ implemented it, but noone _should_).

@frenic , please let me know if you need this syntax, or otherwise disagree with closing the PR.

We should have explicit guidelines for this though.

frenic:
Yeah, some guidelines or checklists would be nice to have. So we get some kind of understanding why a certain spec is missing and when to contribute or not. The reason I added this was because we had some differences with csstype and types used in an library that was made by hand. So the issue is indirect and I really don't see any other useful reason to add it.

PR #204, opened 2018-04-10
Add `justify-all` to text-align

frenic:
No major browsers supports it yet but I'll give it a shot too see what you guys think.

https://www.w3.org/TR/css-text-3/#text-align-property

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6689,7 +6689,7 @@
     "status": "standard"
   },
   "text-align": {
-    "syntax": "start | end | left | right | center | justify | match-parent",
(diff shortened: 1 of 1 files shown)


frenic:
> This something we should decide upon. In general I would say "when the feature is on a standards track and is implemented somewhere" for the earliest point we should think about documenting a standards feature on MDN.

https://github.com/mdn/data/pull/201#issuecomment-381663033

I agree and I can't find any proof that this is implemented somewhere so feel free to close this.

chrisdavidmills:
OK, closing for now. We can always reopen later on.

PR #206, opened 2018-04-13
Add `overflow-anchor`

frenic:
https://drafts.csswg.org/css-scroll-anchoring/#exclusion-api
https://github.com/frenic/csstype/issues/34

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5945,6 +5945,21 @@
     "order": "uniqueOrder",
     "status": "standard"
   },
+  "overflow-anchor": {
+    "syntax": "auto | none",
+    "media": "visual",
+    "inherited": false,
+    "animationType": "discrete",
+    "percentages": "no",
+    "groups": [
+      "CSS Scroll Anchoring"
+    ],
+    "initial": "auto",
+    "appliesto": "allElements",
+    "computed": "asSpecified",
+    "order": "perGrammar",
+    "status": "experimental"
+  },
   "overflow-clip-box": {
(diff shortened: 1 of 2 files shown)


css/properties.json
```diff
@@ -5945,6 +5945,21 @@
...
     "status": "standard"
   },
+  "overflow-anchor": {
+    "syntax": "auto | none",
+    "media": "visual",
+    "inherited": false,
+    "animationType": "discrete",
+    "percentages": "no",
+    "groups": [
+      "CSS Box Model"
```
wbamberg: I'm not sure CSS Box Model is right for this, is it? If "group" should map to the specification then perhaps this should be something like "CSS Scroll Anchoring"?

css/properties.json
```diff
@@ -5945,6 +5945,21 @@
...
+    ],
+    "initial": "auto",
+    "appliesto": "allElements",
+    "computed": "asSpecified",
+    "order": "perGrammar",
+    "status": "standard"
```
wbamberg: Since this is an Editor's Draft and is only implemented by one browser engine AFAIK (https://caniuse.com/#feat=css-overflow-anchor) I think experimental might be more appropriate here? Our guidance on this should be clearer though.

css/properties.json
frenic: Thanks. I felt it was a little bit off too. But wasn't confident enough to create a new one. 😸

css/properties.json
frenic: You're right. Fixed.

PR #207, opened 2018-04-13
Add new overflow properties

frenic:
Adds
- `max-lines`
- `block-overflow`
 - `line-clamp` shorthand for the above
- `overflow-block`
- `overflow-inline`

https://drafts.csswg.org/css-overflow-3/
https://github.com/frenic/csstype/issues/34

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -2054,7 +2054,7 @@
     "animationType": "discrete",
     "percentages": "no",
     "groups": [
-      "CSS Box Model"
+      "CSS Overflow"
     ],
     "initial": "clip",
     "appliesto": "blockContainers",
@@ -4903,7 +4903,7 @@
     "animationType": "integer",
     "percentages": "no",
     "groups": [
-      "CSS Box Model"
+      "CSS Overflow"
     ],
     "initial": "none",
     "appliesto": "blockContainersExceptMultiColumnContainers",
@@ -5536,7 +5536,7 @@
     "animationType": "integer",
     "percentages": "no",
     "groups": [
-      "CSS Box Model"
+      "CSS Overflow"
     ],
     "initial": "none",
     "appliesto": "blockContainersExceptMultiColumnContainers",
@@ -5997,7 +5997,7 @@
     "animationType": "discrete",
     "percentages": "no",
     "groups": [
-      "CSS Box Model"
+      "CSS Overflow"
     ],
     "initial": "visible",
     "appliesto": "nonReplacedBlockAndInlineBlockElements",
@@ -6027,7 +6027,7 @@
     "animationType": "discrete",
     "percentages": "no",
     "groups": [
-      "CSS Box Model"
+      "CSS Overflow"
     ],
     "initial": "auto",
     "appliesto": "nonReplacedBlockAndInlineBlockElements",
@@ -6057,7 +6057,7 @@
     "animationType": "discrete",
     "percentages": "no",
     "groups": [
-      "CSS Box Model"
+      "CSS Overflow"
     ],
     "initial": "auto",
     "appliesto": "nonReplacedBlockAndInlineBlockElements",
@@ -6087,7 +6087,7 @@
     "animationType": "discrete",
     "percentages": "no",
     "groups": [
-      "CSS Box Model"
+      "CSS Overflow"
     ],
     "initial": "visible",
     "appliesto": "nonReplacedBlockAndInlineBlockElements",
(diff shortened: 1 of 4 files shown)


css/properties.json
```diff
@@ -2047,6 +2047,21 @@
...
     "status": "standard"
   },
+  "block-overflow": {
+    "syntax": "clip | ellipsis | <string>",
+    "media": "visual",
+    "inherited": true,
+    "animationType": "no",
+    "percentages": "no",
+    "groups": [
+      "CSS Box Model"
```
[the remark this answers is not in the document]
frenic: Yeah, I was a bit confused about this. Thanks for clearing it out.

css/properties.json
```diff
@@ -2047,6 +2047,21 @@
...
+    ],
+    "initial": "clip",
+    "appliesto": "blockContainers",
+    "computed": "asSpecified",
+    "order": "perGrammar",
+    "status": "standard"
```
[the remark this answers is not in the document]
frenic: Definitely 😄

css/properties.json
```diff
@@ -5975,6 +6020,21 @@
     "order": "perGrammar",
     "status": "experimental"
   },
+  "overflow-block": {
+    "syntax": "<‘overflow’>",
```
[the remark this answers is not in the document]
frenic: Wow, nice catch! Was probably due to copy paste.

css/properties.json
```diff
@@ -5975,6 +6020,21 @@
...
+    "media": "visual",
+    "inherited": false,
+    "animationType": "discrete",
+    "percentages": "no",
+    "groups": [
+      "CSS Box Model"
+    ],
+    "initial": "auto",
+    "appliesto": "nonReplacedBlockAndInlineBlockElements",
```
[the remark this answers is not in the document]
frenic: Seems like it. Changed it for `overflow-[x|y]` too.

frenic:
Fixed your comments @wbamberg. Thanks for your feedback.

css/properties.schema.json
```diff
@@ -175,6 +176,8 @@
         "beforeAndAfterPseudos",
         "blockContainerElements",
         "blockContainers",
+        "blockContainersFlexContainersGridContainers",
```
wbamberg: Nit: this should go after the two `blockContainersExcept*` lines, to keep alpha ordering.

l10n/css.json
```diff
@@ -451,6 +451,12 @@
     "ja": "ブロックコンテナ",
     "ru": "блочные контейнеры"
   },
+  "blockContainersFlexContainersGridContainers": {
+    "en-US": "block-containers, flex containers, and grid containers"
```
wbamberg: Nit: this should have a capital initial:

"Block-containers, flex containers, and grid containers"

css/properties.schema.json
```diff
@@ -26,6 +26,7 @@
         "integer",
         "length",
         "lpc",
+        "no",
```
wbamberg: Oh, also: it seems that where the spec has "Animatable: no", we use "discrete". See for example:
https://developer.mozilla.org/en-US/docs/Web/CSS/align-self#Specifications
https://drafts.csswg.org/css-align-3/#propdef-align-self

wbamberg: I don't know why we do this. "no" seems much clearer, and is what the specs actually say. But there might be a reason I don't understand that we do this.

l10n/css.json
frenic: This is hard to tell of your own. At first glance it looks like half of them are capitalized and the other half isn't. It guess it maybe depends on where it's used?

css/properties.schema.json
frenic: Oh, thank you! I've wondered what this means. You didn't have the full answer but at least you didn't know either. 😁

l10n/css.json
wbamberg: You're right, this is just inconsistent for no good reason that I can see. It is used in the "Applies to" piece of the blue box, in places like https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius#Specifications. As far as I know this is the only place it is used in MDN, although it might be used in other applications.

In the MDN case, it seems to me that capitalization would be better, but most cases actually seem to not capitalize.

I think it's best to just do what you like here, and we will fix it up later :).

frenic: Caps it is then 😄

frenic:
Just noticed that you don't map the computed properties for `background-position` either. It should be the same for `overflow` I guess.

wbamberg:
> I fixed the overflow. Should the computed property be changed to ['overflow-x', 'overflow-y'] as well?

I think probably yes, but I'm not sure. But let's merge what we have here, we can't just keep expanding this PR :).

PR #209, opened 2018-04-19
Fix `-webkit-mask` and `-webkit-mask-*`

frenic:
Noticed that `-webkit-mask` seemed a bit off. Tried my best to refactor it myself but I could much likely do everything wrong here. Documentation is limited.


frenic:
Forgot to mention that I used this source of information:

- https://developer.mozilla.org/en-US/docs/Web/CSS/mask
  >  The prefixed property can be used with SVG and HTML with a slightly different syntax, which allows setting the non-standard -webkit-mask-attachment property.
- https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-mask-attachment
- https://developer.mozilla.org/en-US/docs/Web/CSS/mask-clip
- https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-mask-composite
- https://developer.mozilla.org/en-US/docs/Web/CSS/mask-image
- https://developer.mozilla.org/en-US/docs/Web/CSS/mask-origin
  >  The margin-box value is unsupported.
- https://developer.mozilla.org/en-US/docs/Web/CSS/mask-position
- https://developer.mozilla.org/en-US/docs/Web/CSS/mask-repeat

css/syntaxes.json
```diff
@@ -321,7 +321,7 @@
     "syntax": "linear-gradient( [ <angle> | to <side-or-corner> ]? , <color-stop-list> )"
   },
   "mask-layer": {
-    "syntax": "<mask-reference> || <position> [ / <bg-size> ]? || <repeat-style> || <geometry-box> || [ <geometry-box> | no-clip ] || <compositing-operator> || <masking-mode>"
```
lahmatiy: Currently spec claims that `<geometry-box>` should exists in syntax, see https://drafts.fxtf.org/css-masking-1/#the-mask

css/properties.json
```diff
@@ -1282,7 +1282,7 @@
     "status": "nonstandard"
   },
   "-webkit-mask-origin": {
-    "syntax": "[ padding | border | content ] [, [ border | padding | content ] ]*",
+    "syntax": "[ <box> | fill-box | stroke-box | view-box | no-clip ]#",
```
lahmatiy: What is your source of the syntax?
`-webkit-mask-origin` is not a standard property, and I can only find something like http://css-infos.net/property/-webkit-mask-origin

css/properties.json
```diff
@@ -1267,7 +1267,7 @@
     "status": "nonstandard"
   },
   "-webkit-mask-image": {
-    "syntax": "<mask-image> [, <mask-image> ]*",
+    "syntax": "<'mask-image'>",
```
lahmatiy: `<'mask-image'>` is not the same as `<mask-image>`. Looks like `<mask-image>` is similar, but `<'mask-image'>` can change in future according to the [spec](https://drafts.fxtf.org/css-masking-1/#the-mask-image). It shouldn't affect old syntaxes.

css/properties.json
```diff
@@ -1237,7 +1237,7 @@
     "status": "nonstandard"
   },
   "-webkit-mask-clip": {
-    "syntax": "[ border | border-box | padding | padding-box | content | content-box | text ]#",
+    "syntax": "<'mask-clip'>",
```
lahmatiy: `<-webkit-mask-clip>` is not the same as `<'mask-clip'>`. When I searched info about this property, I found something like this https://gist.github.com/webtobesocial/a27380c066a03e6c87b2#-webkit-mask-clip (and I remember a source about`*-box` values, can't found at the moment).

css/properties.json
```diff
@@ -1193,7 +1193,7 @@
     "status": "nonstandard"
   },
   "-webkit-mask": {
-    "syntax": "<mask-image> [ <'-webkit-mask-repeat'> || <'-webkit-mask-attachment'> || <'-webkit-mask-position'> || <'-webkit-mask-origin'> || <'-webkit-mask-clip'> ]*",
+    "syntax": "[ <mask-reference> || <position> [ / <bg-size> ]? || <repeat-style> || [ <geometry-box> | no-clip ] || <composite-style> || <attachment> ]#",
```
lahmatiy: `-webkit-mask` is not standard and current syntax is more safe. Are you sure a new syntax is relevant to `-webkit-*`  properties?

frenic: Thanks for your feedback. The reason I noticed it was a bit off was firstly because `<mask-image>` doesn't refer to anything. I get warnings for all missing data types when I generate the typings in csstype. What would you suggest it refers to?

Secondly, several of the `<'-webkit-*'>` in the current syntax accepts comma separated values and my suspicions is that `-webkit-mask` doesn't accept values like `url(...) scroll, fixed, padding, content` like this syntax is telling me that I can. Or have I missed somthing?

css/syntaxes.json
frenic: But it still contains `<geometry-box>`. The current syntax if `<geometry-box> || [ <geometry-box> | no-clip ]`. Isn't `<geometry-box> | no-clip` a shortening of that?

lahmatiy: `a || b` means one or more options in any order, so valid values are `a`, `b`, `a b` and `b a`.
`a | b` means one of options, so valid values are `a` and `b`.
See [Component value combinators](https://www.w3.org/TR/css-values-3/#component-combinators) for more details.
There are two entries of `<geometry-box>` in the syntax per [spec](https://drafts.fxtf.org/css-masking-1/#typedef-mask-layer). It means that one or two `<geometry-box>` can be specified in a value.

frenic: > There are two entries of <geometry-box> in the syntax per spec. It means that one or two <geometry-box> can be specified in a value.

Ah, I see. Missed that. Thanks

css/properties.json
lahmatiy: > The reason I noticed it was a bit off was firstly because `<mask-image>` doesn't refer to anything. I get warnings for all missing data types when I generate the typings in csstype. What would you suggest it refers to?

You're right, `<mask-image>` is not defined in `syntax.json`. As I found I missed this issue, because CSSTree has [a patch where this syntax is defined](https://github.com/csstree/csstree/blob/master/data/patch.json#L636-L639). I suppose we should add such syntaxes to `syntax.json`.

On the other hand, `-webkit-mask-*` properties are not standard and we should mitigate influence of it to syntaxes from a specs. For instance `<mask-image>` can be used by a spec later.

> Secondly, several of the `<'-webkit-*'>` in the current syntax accepts comma separated values and my suspicions is that -webkit-mask doesn't accept values like url(...) scroll, fixed, padding, content like this syntax is telling me that I can. Or have I missed somthing?

Agree, current syntax is not valid by the reasons you're mentioning. So this syntax should be reworked. Unfortunately, it is too hard to found a complete information about `-webkit-mask` properties group syntax. I found an [interesting gist](https://gist.github.com/webtobesocial/a27380c066a03e6c87b2#-webkit-mask) – may be it can help.

frenic:
@lahmatiy fixed it a bit. I tried to combine different values and this is what I can up with in latest Chrome. You can see for your self in the image below.

image

`-webkit-mask-attachment` doesn't exist in Chrome so I can't verify that `-webkit-mask` is a shorthand for that property as well. But since it seems to be a Safari-only property maybe it should be described in the `-webkit-mask` syntax at all? Just the longhand property.

I also added the missing `-webkit-mask-size` property.

frenic:
@lahmatiy How should we proceed? Am I on the right track here after my latest changes or should we just fix the missing `<mask-image>` for now? I understand there may be different implementations of `-webpack-mask` that complicates it.

frenic:
Somebody? Anything on this?

wbamberg:
Sorry @frenic , it's been a busy couple of weeks. I will take a look next week.

wbamberg:
I've finally taken a bit of time to dig into this. Apologies for any misunderstandings, I'm still learning.

First, from an "MDN docs" perspective: as a general rule, I think we're less concerned with nonstandard syntaxes than with the standard ones. There aren't MDN docs pages for any of these properties except `-webkit-mask-attachment`, so that is the only one for which MDN pages will use the formal syntax. We might imagine at some point in the future formally describing the differences between, say, `mask` and `-webit-mask`, but we don't do that at the moment, and keeping up with the standard-track stuff keeps us busy enough :). Also, as you discussed, it can be hard to find out what the facts are for proprietary syntaxes... ideally I'd love it if we could get review from the relevant browser vendors for nonstandard syntaxes.

The changes you've made to `-webkit-mask-attachment`, `-webkit-mask-composite`, `-webkit-mask-image`, and `-webkit-mask-repeat` seem straightforwardly fine -  as far as I know this:

```
<attachment> [, <attachment> ]*
```

is just equivalent to this:

```
<attachment>#
```

(if it isn't I'd be glad to learn why)

I agree with you that the current syntax definition for `-webkit-mask` has to be wrong. As I understand it, with `mask`, you get to create one or more mask layers, where each layer includes the image plus optionally other stuff relating to that image, like repeat, position, and origin. This is very much like the way `background` is specified of course. So it doesn't make sense to have one image plus "zero or more" values for repeat, position, origin &c, which is what the current syntax seems to be saying:

```
<mask-image> [ <'-webkit-mask-repeat'> || <'-webkit-mask-attachment'> || <'-webkit-mask-position'> || <'-webkit-mask-origin'> || <'-webkit-mask-clip'> ]*
```

What you have here is much closer to the standard syntax, and definitely looks better than what there was before. I haven't tried to verify all the details of where the syntax diverges. But I would be happy to r+ this as it is, since it does seem to be an improvement.

As for adding `<mask-image>`...

* `"mask-image"` is listed in properties.json with the syntax `<mask-reference>#`.

* As far as I can tell, the only places where a syntax refers to `<mask-image>` are `"-webkit-mask"` and `"-webkit-mask-image"`.

* In particular, the standard `"mask"` property doesn't refer to `"mask-image"` - instead it uses `<mask-layer>#`, and `<mask-layer>` then uses `<mask-reference>` directly.

So instead of adding `<mask-image>` to syntaxes.json, your PR changes the definitions of `-webkit-mask` and `-webkit-mask-image` to use `<mask-reference>` directly. I think this seems like a better approach, as it's closer to the way the standard properties are defined (and as we said, it fixes a bug in the `-webkit-mask` syntax). And it appears to me as if this would enable @lahmatiy to remove the [patch in css-tree](https://github.com/csstree/csstree/blob/master/data/patch.json#L636-L639).

So all that said, I think I'm r+ on this patch :). But I'd also like to know what @lahmatiy thinks, especially on the matter of `<mask-image>`.

lahmatiy:
> Unless you're forced to use a white space between the , and `<attachment>` for some reason. But it doesn't seem to be that way in Chrome.

FYI, whitespaces and comments around a comma are optional, like in many other cases.

> like to know what @lahmatiy thinks

Sorry, I have no time to review this PR. Those syntaxes are quite complicated in various aspects and require a research. Currently I'm working on a new syntax matching approach in CSSTree. I believe it can help to validate those (and other) syntaxes in one way or another. Don't await for me in case you want to push this PR further. I will back with fixes or suggestions later if they appear ;)

PR #211, opened 2018-04-20
Add `-webkit-appearance`

frenic:
I took all the values from latest Chrome.


frenic:
I compared the values I found in Chrome with the ones [implemented in Safari](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariCSSRef/Articles/StandardCSSProperties.html
). If i filter the values that are common between these two the syntax would be:

`none | button | button-bevel | caret | checkbox | listbox | listitem | media-mute-button | media-play-button | media-slider | media-sliderthumb | menulist | menulist-button | menulist-text | menulist-textfield | push-button | radio | searchfield | searchfield-cancel-button | slider-horizontal | slider-vertical | sliderthumb-horizontal | sliderthumb-vertical | square-button | textarea | textfield`

Would this be more appropriate to use instead?

frenic:
Or should it be a mix of values used by both Chrome and Safari?

frenic:
Nothing? C'mon this property is fairly used. It should exist in some level even if there's some complexity to it.

wbamberg:
I've verified that this PR contains:

* all the Safari values from https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariCSSRef/Articles/StandardCSSProperties.html

* all the Chrome values from https://developer.mozilla.org//en-US/docs/Web/CSS/-moz-appearance#Values

* ...no values that are not in one of those 2 lists.

CSS Tricks lists a number of [`scroll*` values](https://css-tricks.com/almanac/properties/a/appearance/) which are missing here. Is there a reason to omit them?

> I think it should be a mix if Chrome and Safari, if we talk about -webkit-appearance.

I guess the problem with this is that a given value isn't necessarily valid, depending on which browser you're talking about.

I hesitate to suggest this, but it feels like what we *should* do in circumstances like this is model the different syntaxes in the data, so rather than just having `syntax` be a string it could be allowed to be, say, an array of objects, like:

```
"syntax": [
    {
        "runtime": "chrome"
        "values": "chrome values..."
    },
    {
        "runtime": "safari"
        "values": "safari values..."
    }
]
```

That way the user of the data could get correct data for the particular runtime.

But maybe this isn't worth doing at the moment.

lahmatiy:
@wbamberg I like you suggestion, it's on the right way. However it's too complicated for now and doesn't solve other issues, like various browser/version support for standard syntaxes (e.g. values for `display` have so different support), new/old syntaxes by spec (e.g. latest changes in rgb/rgba/hsl/hsla when a spec introduced syntaxes with no comma and makes rgba/hsla just an alias for rgb/hsl) and so on. Let's move this discussion to a separate issue.

wbamberg:
Oh, to be clear, I'd only imagine this for non-standard items. When there is a standard we should have only the most up-to-date version in mdn/data (this means I was probably wrong about https://github.com/mdn/data/pull/204) and use [browser-compat-data](https://github.com/mdn/browser-compat-data) to document the level of browser support and notable incompatibilities, etc.

But anyway, yes, this should have more discussion outside this PR :).

frenic:
I like the idea of defining different nonstandard syntaxes for different vendors! 👍 

> CSS Tricks lists a number of scroll* values which are missing here. Is there a reason to omit them?

Well I didn't include the values that where removed in Safari 4.0. From [the documentation](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariCSSRef/Articles/StandardCSSProperties.html) I found this:

> The following constants are unsupported in Safari 4.0:
```scrollbarbutton-down, scrollbarbutton-left, scrollbarbutton-right, scrollbarbutton-up, scrollbargripper-horizontal, scrollbargripper-vertical, scrollbarthumb-horizontal, scrollbarthumb-vertical, scrollbartrack-horizontal, scrollbartrack-vertical```

Do you think they should be included anyway?

wbamberg:
> Well I didn't include the values that where removed in Safari 4.0. From the documentation I found this:

But are they in Chrome?

frenic:
Not in latest Chrome. Found [this](https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/css/CSSValueKeywords.json5?l=641) to somehow prove it.

frenic:
Thank you 😸

PR #219, opened 2018-05-02
Update `text-align` syntax

mfluehr:
This adds the `justify-all` keyword, as per https://drafts.csswg.org/css-text-3/#valdef-text-align-justify-all.


mfluehr:
Never mind, I saw #204.

PR #221, opened 2018-05-06
Fix a syntax errors for @font-face

lahmatiy:
I'm working on a new syntax parser and validator in CSSTree. Now it parses a complicated syntaxes and I found an error in `@font-face` syntaxes.

Btw, syntaxes in `css/at-rule.json` have several issues (see #222)

### DIFF
--- a/css/at-rules.json
+++ b/css/at-rules.json
@@ -120,7 +120,7 @@
     "status": "nonstandard"
   },
   "@font-face": {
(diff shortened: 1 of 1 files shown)


wbamberg:
Yes, I'll have a look today. Hoping to make a release today as well, so if there are any other PRs you'd really like to see, let me know.

PR #224, opened 2018-05-07
Fix data for line-height-step

xfq:
Ref: https://github.com/w3c/csswg-drafts/commit/b752cd9fd688fd9dd5d846f003f197bfd1792e54

URL of the relevant spec section: https://drafts.csswg.org/css-rhythm/#line-height-step


lahmatiy:
LGTM

PR #225, opened 2018-05-09
Set initial values for <'-ms-scrollbar-*-color'> properties

lahmatiy:
Used https://github.com/w3c/csswg-drafts/issues/1956 as a source of truth.
Unfortunately, there is no an initial value for `-ms-scrollbar-base-color` found.

//cc @jameshkramer

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -338,7 +338,7 @@
     "groups": [
       "Microsoft Extensions"
     ],
-    "initial": "dependsOnUserAgent",
+    "initial": "ButtonText",
     "appliesto": "allElements",
     "computed": "asSpecified",
     "order": "uniqueOrder",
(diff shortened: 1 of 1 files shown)


lahmatiy:
I suppose we can update initial value for `-ms-scrollbar-base-color` later.

jameshkramer:
The original MSDN article for -ms-scrollbar-base-color says "depends on user-agent" for the initial value (I've tried to attach the document). @lahmatiy and I have not had any success at experimentally determining initial values for the -ms-scrollbar... properties ( see #177 ), so I don't see that we will obtain the initial value for this one.

[scrollbarBaseColor.docx](https://github.com/mdn/data/files/1993504/scrollbarBaseColor.docx)

//cc: @wbamberg @erikadoyle

erikadoyle:
I tried looking for it, but couldn't find any hard-coded values. Seems to get determined at runtime, but I'm not exactly sure when (and where/how).

PR #226, opened 2018-05-09
Update (fix) syntaxes for <rgb>, <rgba>, <hsl> and <hsla>

lahmatiy:
Syntax for `<rgb>`, `<rgba>`, `<hsl>` and `<hsla>` were updated in [CSS Color Module Level 4](https://drafts.csswg.org/css-color/).
Previously syntaxes were broken since group for arguments were missed (i.e. `rgb( [case1] | [case2] )` must be `rgb( [ [case1] | [case2] ] )`).

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -258,10 +258,10 @@
     "syntax": "[ historical-ligatures | no-historical-ligatures ]"
   },
   "hsl()": {
-    "syntax": "hsl( [ <hue> <percentage> <percentage> [ / <alpha-value> ]? ] | [ <hue>, <percentage>, <percentage>, <alpha-value>? ] )"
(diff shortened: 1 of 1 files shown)


deedubbu:
Seems like this change is causing [csstree/validator](https://github.com/csstree/validator) to suddenly report the following line as "Invalid value for `background-color`". 

`background-color: rgba(0, 0, 0, .12);`

Please correct me if I'm overlooking a syntax error but it seems valid to me. Wasn't sure where to open an issue to track but I investigated and determined if I forced mdn-data to be 1.1.2 instead of 1.1.3 I don't get the invalid value error.

lahmatiy:
@deedubbu Yep, the current version of CSSTree can't handle such form of syntax correctly. However, CSSTree's syntax matching was reworked and now it much better (see https://twitter.com/csstree/status/997251928039022592). Release of CSSTree is coming soon with updates of validator and related plugins.
Sorry for inconvenience.

deedubbu:
@lahmatiy Thank you! I've confirmed that the latest versions fix the issue I was seeing.

Issue #222, opened 2018-05-06
Improvements on at-rule syntax description


lahmatiy:
There are some syntax description issues in `css/at-rule.json`:
- `@charset \"<charset>\";` is quite incorrect, should be `@charset <string> ;`
- `@counter-style` has a description of block, but declaration refers to wrong types (i.e. `system: <counter-system>;` but `system` key in `descriptors`)
- `@font-face` had a problem with wrong reference like in `@counter-style` syntax, but I fixed it in #221
- `<group-rule-body>` is wrong for the block of `@viewport` (should be a list of descriptors)

I think we need to change the description of `@counter-style`, `@font-face` and `@viewport`, since those at-rules have a limited list of descriptors and block syntax can be generated from a `descriptors` section. The rest at-rule descriptions should be revisited too.
I can bring a PR later if there is no blockers for changes (like MDN pages).

skyclouds2001:
the onlt remain ones are `@charset` and `@counter-style`

PR #237, opened 2018-05-25
Update @font-face for CSS Fonts Level 4

wbamberg:
This updates the font property descriptors `font-style`, `font-stretch`, `font-weight`, for CSS Fonts Level 4: https://drafts.csswg.org/css-fonts-4/#font-prop-desc.

@lahmatiy , since you recently rewrote this syntax, I'd be very happy to get your review on this change.

(Also, what do you think of removing the newline characters? I'd prefer for the syntax not to try to determine the presentation, but leave that up to the application.)

### DIFF
--- a/css/at-rules.json
+++ b/css/at-rules.json
@@ -120,7 +120,7 @@
     "status": "nonstandard"
   },
   "@font-face": {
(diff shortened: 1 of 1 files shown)


lahmatiy:
@wbamberg I think you should change a syntaxes in `descriptors` section, instead of at-rule syntax (see #222).
```
<font-weight-absolute> <font-weight-absolute>?
```
might be
```
<font-weight-absolute>{1,2}
```
Another case
```
[ <angle> | <angle> <angle> ] ?
```
Multipliers must stick to a group.

wbamberg:
@lahmatiy , does this look better? :)

lahmatiy:
@wbamberg Yep, that's better. But looks like `<font-stretch-absolute>` and `<font-weight-absolute>` syntaxes are not defined.

wbamberg:
>  looks like <font-stretch-absolute> and <font-weight-absolute> syntaxes are not defined.

They should be in https://github.com/mdn/data/pull/235 and https://github.com/mdn/data/pull/234.

Issue #239, opened 2018-06-01
Update CSS properties data to remove multiple groups


wbamberg:
Phase 1 of https://github.com/mdn/data/issues/238: go through CSS property data and update it so every property only belongs to a single group. This would fix the immediate sidebar problem, without needing a schema update.

wbamberg:
@jmswisher , can I add this to the sprint? I don't see "Sprint 4" in the list of available milestones for this issue...

wbamberg:
-> https://github.com/mdn/data/pull/243

PR #243, opened 2018-06-05
Make standard CSS properties only belong to one group

wbamberg:
This is a partial resolution to https://github.com/mdn/data/issues/239.

It updates data for standard CSS properties to make sure that they only belong to one "group", that being the most recent spec that defines them. This should be enough to make sidebars more coherent for pages like `column-gap`.

This is a little dodgy, I suppose, since the "most recent spec" is generally less mature than the others (e.g. `break-after` is still defined in CSS Regions, with the note:

> This section is also defined in [CSS3-BREAK]. If that specification moves to last call with the region values, the section here can be replaced by a reference.

(https://drafts.csswg.org/css-regions-1/#region-flow-break)

But I guess that realistically, it's safe to consider that `break-after` is now defined in the CSS Fragmentation spec.

This PR does not update data for various nonstandard properties:

```
box-flex-group
box-ordinal-group
box-flex
-moz-appearance
box-lines
box-orient
box-pack
box-direction
box-align
```

@rachelandrew , please let me know if this looks good to you.

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -3295,9 +3295,7 @@
     "animationType": "discrete",
     "percentages": "no",
     "groups": [
-      "CSS Columns",
-      "CSS Fragmentation",
-      "CSS Regions"
+      "CSS Fragmentation"
     ],
     "initial": "auto",
     "appliesto": "blockLevelElements",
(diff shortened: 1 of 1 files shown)


rachelandrew:
Looks good and I agree re break-after, that should be considered part of fragmentation.

Issue #172, opened 2018-02-07
Validate syntaxes


teoli2003:
Hi!

Properties and selector syntaxes make use of lower-level syntaxes.

Unless they are "basic" types (that has their own page on MDN), these lower-level syntax are defined in syntaxes.json (and so on, recursively)

It is fastidious and error-prone to be sure not to forget one of these lower-level syntaxes definition. It is also complex for the reviewers to spot them.

We should write a test, run by ci that detect this.

lahmatiy:
I have long wanted to add syntax validation using CSSTree into test script. Currently CSSTree supports [CSS Values and Units module](https://drafts.csswg.org/css-values/) grammar. However some syntaxes (e.g. selectors, at-rules etc) has extended grammar (i.e. uses tokens that are not defined by V&U module) and can't be parsed by CSSTree at the moment. There are two options:
- improve CSSTree to be able process all the syntaxes, but that's need a time – don't know when it happen
- add a validation but use a blacklist for a `bad` syntaxes – that's not an ideal solution, but most syntaxes can be validated right now

What do you think? Are there other options how to do this?

wbamberg:
I think this would be great, and had been wondering the same thing. My preference would be your second option. Even simpler, could we just validate properties (and their dependent syntax)? Just that would be really useful.

PR #244, opened 2018-06-14
Validate syntaxes

abaco:
I made a test to validate the syntaxes in `properties.json` and `syntaxes.json`, as suggested in #172.

As of now the test fails as there are several types that are neither in `syntaxes.json` nor in `types.json`:
```
any-value
attr-fallback
attr-name
custom-property-name
declaration
declaration-list
declaration-value
dimension
function-token
hash-token
hex-color
ident-token
name-repeat
path()
positive-integer
string-token
x
y
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -92,6 +92,16 @@
       "integrity": "sha1-p9BVi9icQveV3UIyj3QIMcpTvCU=",
       "dev": true
     },
+    "css-tree": {
+      "version": "1.0.0-alpha.29",
+      "resolved": "https://registry.npmjs.org/css-tree/-/css-tree-1.0.0-alpha.29.tgz",
+      "integrity": "sha512-sRNb1XydwkW9IOci6iB2xmy8IGCj6r/fr+JWitvJ2JxQRPzN3T4AGGVWCMlVmVwM1gtgALJRmGIlWv5ppnGGkg==",
+      "dev": true,
+      "requires": {
+        "mdn-data": "~1.1.0",
+        "source-map": "^0.5.3"
+      }
+    },
     "escape-string-regexp": {
       "version": "1.0.5",
       "resolved": "https://registry.npmjs.org/escape-string-regexp/-/escape-string-regexp-1.0.5.tgz",
@@ -150,6 +160,18 @@
       "version": "2.1.0",
       "resolved": "https://registry.npmjs.org/leven/-/leven-2.1.0.tgz",
       "integrity": "sha1-wuep93IJTe6dNCAq6KzORoeHVYA=",
+      "dev": true
+    },
+    "mdn-data": {
+      "version": "1.1.4",
+      "resolved": "https://registry.npmjs.org/mdn-data/-/mdn-data-1.1.4.tgz",
+      "integrity": "sha512-FSYbp3lyKjyj3E7fMl6rYvUdX0FBXaluGqlFoYESWQlyUTq8R+wp0rkFxoYFqZlHCvsUXGjyJmLQSnXToYhOSA==",
+      "dev": true
+    },
+    "source-map": {
+      "version": "0.5.7",
+      "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.7.tgz",
+      "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w=",
       "dev": true
     },
     "supports-color": {
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
   },
   "scripts": {
     "lint": "node test/lint",
-    "travis": "npm test",
-    "test": "node test/test"
+    "validate-syntaxes": "node test/validate-syntaxes",
+    "test": "npm run lint && npm run validate-syntaxes",
+    "travis": "npm test"
   }
 }
--- a/package.json
+++ b/package.json
@@ -26,11 +26,12 @@
   "homepage": "https://developer.mozilla.org",
   "devDependencies": {
     "ajv": "^5.0.1",
-    "better-ajv-errors": "^0.5.1"
+    "better-ajv-errors": "^0.5.1",
+    "css-tree": "^1.0.0-alpha.29"
   },
   "scripts": {
     "lint": "node test/lint",
     "travis": "npm test",
-    "test": "npm run lint"
+    "test": "node test/test"
   }
 }
--- a/test/test.js
+++ b/test/test.js
@@ -1,0 +1,2 @@
+require('./lint');
+require('./validate-syntaxes');
--- a/test/test.js
+++ b/test/test.js
@@ -1,2 +1,0 @@
-require('./lint');
-require('./validate-syntaxes');
--- a/test/validate-syntaxes.js
+++ b/test/validate-syntaxes.js
@@ -2,9 +2,9 @@ var properties = require('../css/properties');
 var syntaxes = require('../css/syntaxes');
 var basicTypes = require('../css/types');
 var parseGrammar = require('css-tree').grammar.parse;
+var walkGrammar = require('css-tree').grammar.walk;
 
 var hasErrors = false;
-var itemsHaveErrors;
 
 function isBasicType(name) { return basicTypes.hasOwnProperty(name); }
 function isSyntax(name) { return syntaxes.hasOwnProperty(name); }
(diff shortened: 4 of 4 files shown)


abaco:
As far as I understand the types in `properties.json` can refer to other properties, but those in `syntaxes.json` can't. I fixed the test accordingly. Indeed, [`<shape>`](https://developer.mozilla.org/en-US/docs/Web/CSS/shape) has subtypes `<top>`, `<right>`, `<bottom>`, `<left>`, but those are synonims to the basic type `<length>`, not the properties with the same names.

package-lock.json
```diff
@@ -152,6 +162,18 @@
       "integrity": "sha1-wuep93IJTe6dNCAq6KzORoeHVYA=",
       "dev": true
     },
+    "mdn-data": {
```
connorshea: Will importing mdn-data from npm cause any problems, e.g. if you add a new type in the dataset will the tests start failing until a new release of the npm package is out?

abaco: That `mdn-data` npm package is there because css-tree depends on it, but it's never used. The test loads the local data (e.g. `require('../css/properties')`).

abaco:
I think I was wrong about `syntaxes.json` not refererring to `properties.json`. There's at least one case: the syntax [`inset()`](https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape) refers to the property `border-radius`. I should then revert my last commit bf89a7edea86f7d136329871b2bbd21b28c33a42. Am I right?

I still believe there's a problem with the syntax `shape`. It validates because its subtypes `top`, `bottom`, `left`, and `right` are interpreted as properties, but in fact they should be plain `length`s. That's a separate issue, though.

wbamberg:
@abaco , thanks for doing this! Everyone in Mozilla is at an event this week, but I will take a look next week after I get back.

wbamberg:
@abaco , about tokens, see also https://github.com/mdn/data/pull/169#issuecomment-392992902. I think having them in a separate "tokens.json" might be a good idea. I would be interested to know what @lahmatiy thinks about that too though.

wbamberg:
> I still believe there's a problem with the syntax shape. It validates because its subtypes top, bottom, left, and right are interpreted as properties, but in fact they should be plain lengths.

What makes you say they should be plain lengths? `<shape>` is deprecated, but appears to be defined here: https://www.w3.org/TR/CSS2/visufx.html#value-def-shape, which says:

> the only valid <shape> value is: rect(<top>, <right>, <bottom>, <left>) ... <top>, <right>, <bottom>, and <left> may either have a <length> value or 'auto'.

...so unless this is the wrong reference, then they are not plain lengths. On the other hand, they are also not the same syntax as the [`<top>` etc properties](https://github.com/mdn/data/blob/master/css/properties.json#L7225), since they don't appear to accept a percentage.

Since this is deprecated I don't think it would be a problem to change them to plain lengths though, this seems closer to accurate than using `<'top'>`.

test/validate-syntaxes.js
```diff
@@ -0,0 +1,76 @@
...
+  switch (grammar.type) {
+    case 'Group':
+      return grammar.terms
+        .map(validateGrammar)
+        .reduce(function(a, b) { return a.concat(b); }, []);
+    case 'Multiplier':
+      return validateGrammar(grammar.term);
+    case 'Type':
+      var typeName = grammar.name;
+      if (!(isBasicType(typeName) || isSyntax(typeName))) {
```
wbamberg: So I guess to accommodate the `inset()` case this will also need to look in `properties`?

abaco:
@wbamberg Sorry, I created some confusion, but I think I finally understand how references to syntaxes work :smile: I was misled by `inset()` containing `<border-radius>`: I thought `<foo>` could refer to an item in types.json, syntaxes.json, or properties.json. Then I realized properties must be referred to with quotes, so `<border-radius>` should fail, and it should be changed to `<'border-radius'>`. So I did #246 and updated the validation code.

With that change the validation of [`shape`](https://github.com/mdn/data/blob/master/css/syntaxes.json#L566-L568) also fails, as it should. `<top>`, `<right>`, `<bottom>` and `<left>` can't refer to properties because they lack the quotes. I said they are plain `<length>`s because I read that in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/shape#Syntax), but if that's wrong then we should fix the docs and maybe create a new type like

```
"length-auto": {
  "syntax": "<length> | auto"
}
```

and update `shape` to `rect(<length-auto>, <length-auto>, <length-auto>, <length-auto>)`.

> I think it would be cleaner to separate these roles, and have properties.json only contain properties, and include the types in syntax.json.

I think the way of referring to properties with quotes clearly indicate the roles. The property `border-radius` can take the role of a type when used as `<'border-radius'>`. That's clear enough - once you know it! :smile:

I think it's better to fix the syntaxes that don't validate yet in separate PRs, to avoid merge conflicts. But of course let's discuss the general approaches (like what to do with tokens).

test/validate-syntaxes.js
abaco: The `inset()` case was fixed with #246. We only need to look in `properties` when the `grammar.type` is `'Property'`, which correponds to cases like `<'border-radius'>`.

package.json
```diff
@@ -26,11 +26,12 @@
...
     "ajv": "^5.0.1",
-    "better-ajv-errors": "^0.5.1"
+    "better-ajv-errors": "^0.5.1",
+    "css-tree": "^1.0.0-alpha.29"
   },
   "scripts": {
     "lint": "node test/lint",
     "travis": "npm test",
-    "test": "npm run lint"
+    "test": "node test/test"
```
lahmatiy: It's a bad idea to hide various checkers inside a module. In my opinion it's better to make separate commands:
```
"scripts": {
     "lint": "node test/lint",
     "validate-syntaxes": "node test/validate-syntaxes",
     "test": "npm run lint && npm run validate-syntaxes",
     "travis": "npm test"
}
```

test/test.js
```diff
@@ -0,0 +1,2 @@
+require('./lint');
```
lahmatiy: Better to mix commands on `scripts` level – this file is redundant

test/validate-syntaxes.js
```diff
@@ -0,0 +1,84 @@
...
+      return grammar.terms
+        .map(validateGrammar)
+        .reduce(function(a, b) { return a.concat(b); }, []);
+    case 'Multiplier':
+      // validate the nested grammar
+      return validateGrammar(grammar.term);
+    case 'Type':
+      // validate basic types and non-terminal types, e.g. <length>, <bg-image>
+      var typeName = grammar.name;
+      if (!(isBasicType(typeName) || isSyntax(typeName))) {
```
lahmatiy:
```
!isBasicType(typeName) && !isSyntax(typeName)
```

test/validate-syntaxes.js
```diff
@@ -0,0 +1,84 @@
...
+var hasErrors = false;
+var itemsHaveErrors;
+
+function isBasicType(name) { return basicTypes.hasOwnProperty(name); }
+function isSyntax(name) { return syntaxes.hasOwnProperty(name); }
+function isProperty(name) { return properties.hasOwnProperty(name); }
+
+// Checks that the syntaxes this grammar refers to exist.
+// Returns an array of errors (strings), one for each broken reference.
+function validateGrammar(grammar) {
```
lahmatiy: Please, use csstree's walker instead your own:
```js
var walkGrammar = require('css-tree').grammar.walk;

function validateGrammar(grammar) {
  var errors = [];

  walkGrammar(grammar, function(node) {
    switch (node.type) {
      case 'Type':
        // validate basic types and non-terminal types, e.g. <length>, <bg-image>
        var typeName = node.name;
        if (!isBasicType(typeName) && !isSyntax(typeName))) {
          errors.push('Broken reference: <' + typeName + '>');
        }
        break;

      case 'Property':
        // validate references to properties, e.g. <'background-color'>
        var typeName = node.name;
        if (!isProperty(typeName)) {
          errors.push('Broken reference: <\'' + typeName + '\'>');
        }
        break;
    }
  });

  return errors;
}
```

test/validate-syntaxes.js
```diff
@@ -0,0 +1,84 @@
...
+}
+
+function validateItem(name, item) {
+  try {
+    var syntax = parseGrammar(item.syntax);
+    var errs = validateGrammar(syntax);
+    if (errs.length > 0) {
+      // print validation errors
+      console.log('  ' + name + ':');
+      errs.forEach(function (err) { console.log('    ' + err); });
```
lahmatiy:
```js
console.log('    ' + errs.join('\n    '))
```

test/validate-syntaxes.js
```diff
@@ -0,0 +1,84 @@
...
+      return [];
+    default:
+      return [];
+  }
+}
+
+function validateItem(name, item) {
+  try {
+    var syntax = parseGrammar(item.syntax);
+    var errs = validateGrammar(syntax);
```
lahmatiy: My suggestion to rename `errs` -> `errors`

test/validate-syntaxes.js
```diff
@@ -0,0 +1,84 @@
...
+    if (errs.length > 0) {
+      // print validation errors
+      console.log('  ' + name + ':');
+      errs.forEach(function (err) { console.log('    ' + err); });
+      itemsHaveErrors = true;
+    }
+  }
+  catch (err) {
```
lahmatiy: Should not to be on new line

test/validate-syntaxes.js
```diff
@@ -0,0 +1,84 @@
...
+
+  itemsHaveErrors = false;
+
+  Object.keys(items).forEach(function(key) { validateItem(key, items[key]); });
+
+  if (itemsHaveErrors) {
+    console.log('');
+    hasErrors = true;
+  }
+  else {
```
lahmatiy: Should not to be on new line

test/validate-syntaxes.js
```diff
@@ -0,0 +1,84 @@
...
+
+function validateItems(file, items) {
+  console.log(file);
+
+  itemsHaveErrors = false;
+
+  Object.keys(items).forEach(function(key) { validateItem(key, items[key]); });
+
+  if (itemsHaveErrors) {
+    console.log('');
```
lahmatiy: Redundant

lahmatiy: Redundant

test/validate-syntaxes.js
```diff
@@ -0,0 +1,84 @@
...
+    console.log('  ' + name + ':');
+    console.log('    ' + err.message.replace(/\n/g, '\n    '));
+    itemsHaveErrors = true;
+  }
+}
+
+function validateItems(file, items) {
+  console.log(file);
+
+  itemsHaveErrors = false;
```
lahmatiy: I believe it can be a local variable

lahmatiy:
@wbamberg I would be nice to add eslint (or any other syntax linter) to the project, because each PR related to scripts in JavaScript brings its own code style.

lahmatiy:
@abaco Thank you for contribution!
I thought to add syntax checking with CSSTree by my self. But I was stopped because CSSTree has no definition for "generic" types like `<dimension>` or `<*-token>`. So I decide to wait until CSSTree get such dictionaries, to avoid necessary for exceptions in the code.
I suppose alpha30 (or alpha31) will get those definitions, then we will can to check all the syntaxes. Until that you can add a white list to filter some errors.

abaco:
Alright! So I'll make a whitelist with all the "generic" types, but there are still some syntaxes that shouldn't be whitelisted:

- in `<shape>`: `<top>`, `<right>`, etc. (#248)
- in `<cursor>`; `<x>` and `<y>`
- in `<offset-path>`: `<path()>`
- any other I'm missing?

These won't be recognized by CSSTree even after the update, so they should be fixed independently. It's fine to filter out errors, but I want to make sure I'm not swiping bugs under the carpet.

PR #246, opened 2018-06-19
syntaxes.json: <'border-radius'> instead of <border-radius>

abaco:
In [syntaxes.json](https://github.com/mdn/data/blob/master/css/syntaxes.json#L315) the non-terminal data type `<border-radius>` should be enclosed in quotes, as explained in [Value definition syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#Non-terminal_data_types):

> data types *sharing the same name of a property*, put between quotes


lahmatiy:
LGTM

PR #248, opened 2018-07-02
Fix <shape> syntax

abaco:
As mentioned in #244 the syntax `<shape>` should be fixed.

The [correct syntax](https://www.w3.org/TR/CSS2/visufx.html#value-def-shape) of `<top>`, `<right>`... is `<length> | auto`. The [MND docs](https://developer.mozilla.org/en-US/docs/Web/CSS/shape#Syntax) erroneously state that it's `<length>`.

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -335,6 +335,9 @@
   "leader-type": {
     "syntax": "dotted | solid | space | <string>"
   },
+  "length-auto": {
+    "syntax": "<length> | auto"
+  },
   "length-percentage": {
     "syntax": "<length> | <percentage>"
   },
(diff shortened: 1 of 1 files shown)


abaco:
@lahmatiy 
I didn't know we had to be faithful to the spec, but now that I know I fully agree.

The spec univocally defines top/right/bottom/left. Their *names* may not be the most appropriate, but I don't see why we shouldn't treat them as every other syntax. IMO the most consistent solution is to add them to syntaxes.json (as you did in CSSTree: https://github.com/csstree/csstree/blob/master/data/patch.json#L673-L675).

If the labels top/right/bottom/left are really so bad, that could be fixed in the spec, but it shouldn't stop this project from being consistent to the current spec and to itself. What do you think?

connorshea:
before this gets merged @abaco should probably agree to licensing this under CC-0 :) https://github.com/mdn/data/issues/240

abaco:
@connorshea I've agreed to licensing under CC-0, but lahmatiy's objections are valid so this PR shouldn't be merged.

PR #249, opened 2018-07-04
Add data for scrollbar-*-color

wbamberg:
See https://drafts.csswg.org/css-scrollbars-1/#scrollbar-color-properties

### DIFF
--- a/css/definitions.json
+++ b/css/definitions.json
@@ -42,6 +42,7 @@
       "CSS Ruby",
       "CSS Scroll Anchoring",
       "CSS Scroll Snap",
+      "CSS Scrollbars",
       "CSS Shapes",
       "CSS Speech",
       "CSS Table",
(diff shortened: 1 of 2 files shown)


wbamberg:
Hmm, this is because of https://bugzilla.mozilla.org/show_bug.cgi?id=1460456, which landed support for these two properties in Firefox 62. But looking more closely I think perhaps they are not exposed to Web pages yet, in which case we should probably not document them yet. I've asked for more clarification.

wbamberg:
-> https://bugzilla.mozilla.org/show_bug.cgi?id=1460456#c34, so it is not exposed to the web. I've asked @chrisdavidmills to confirm, but it seems likely to me that we should not document these properties now, so we should not merge this PR.

chrisdavidmills:
I've commented in the bug too, but yes, let's leave this for now. It looks to be preffed off, and not even exposed to the web yet. Let's leave this for now.

PR #250, opened 2018-07-04
Added 'x' as a unit for 'resolution'

wbamberg:
See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1460655
https://drafts.csswg.org/css-values/#resolution


lahmatiy:
LGTM
([background on adding it to the spec](https://github.com/w3c/csswg-drafts/issues/461#issuecomment-298972588))

PR #252, opened 2018-08-07
Add place-items property.

### LINKED ISSUE #228: `place-items` property is missing
https://developer.mozilla.org/en-US/docs/Web/CSS/place-items

connorshea:
Resolves #228. This is my first time doing this sort of thing, but I think I've got it right.

Sources:

- https://drafts.csswg.org/css-align-3/#place-items-property
- Data for similar `place-content` property
- https://developer.mozilla.org/en-US/docs/Web/CSS/place-items

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6835,10 +6835,7 @@
     "syntax": "<'align-items'> <'justify-items'>?",
     "media": "visual",
     "inherited": false,
-    "animationType": [
-      "align-items",
-      "justify-items"
-    ],
+    "animationType": "discrete",
     "percentages": "no",
     "groups": [
       "CSS Box Alignment"
--- a/css/properties.json
+++ b/css/properties.json
@@ -6831,6 +6831,30 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/place-content"
   },
+  "place-items": {
+    "syntax": "<'align-items'> <'justify-items'>?",
+    "media": "visual",
+    "inherited": false,
+    "animationType": [
+      "align-items",
+      "justify-items"
+    ],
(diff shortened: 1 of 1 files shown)


css/properties.json
```diff
@@ -6472,6 +6472,30 @@
     "order": "uniqueOrder",
     "status": "standard"
   },
+  "place-items": {
+    "syntax": "<'align-items'> <'justify-items'>?",
+    "media": "visual",
+    "inherited": false,
+    "animationType": [
```
connorshea: One question: Should I use an array like this for the shorthand property, and if so why does `place-content` not use arrays?

connorshea: If it shouldn't use an array, here's the format I used initially (including this for my future reference if it's ever necessary):
```json
"animationType": "discrete",
"initial": "normal legacy",
"computed": "asSpecified"
```

wbamberg: I'm not sure, and I wish the guidance was clearer here, but I think it's because [the specification for the property](https://drafts.csswg.org/css-align-3/#place-items-property) has:

Animatable: | no
-- | --

...so there's no value in saying "as each of the properties of the shorthand:", it's more direct and simpler just to say "no" (or "discrete" as we put it, which seems confusing but is perhaps driven by the issue linked from https://drafts.csswg.org/css-transitions/#animatable-properties).

Does that sound plausible to you?

connorshea: Sounds reasonable to me, I've updated the PR.

connorshea:
@wbamberg should be good to go now 👍

connorshea:
I noticed master has added MDN URLs for all properties so I rebased and added the MDN URL for place-items :)

PR #259, opened 2018-08-16
Add MDN URLs to selectors data

ddbeck:
This is a continuation of #233. This PR adds MDN URLs to the selectors data and schema.

Two entries didn't get URLs:
- `:host-context` (not clear to me this is even a thing)
- `::-webkit-progress-inner-value` (doesn't seem to have a page)

### DIFF
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -5,7 +5,8 @@
       "Basic Selectors",
       "Selectors"
     ],
-    "status": "standard"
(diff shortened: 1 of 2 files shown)


wbamberg:
> :host-context (not clear to me this is even a thing)

Yes, this looks like an error (reviewed by me :( -> https://github.com/mdn/data/pull/198).

PR #260, opened 2018-08-16
Add MDN URLs to types data

ddbeck:
This is yet another continuation of #233. This PR adds MDN URLs to the CSS types data and schema.

Two entries didn't get URLs:

- `an-plus-b` (not sure what this is referring to)
- `ident` (couldn't find a page)

### DIFF
--- a/css/types.json
+++ b/css/types.json
@@ -9,28 +9,32 @@
     "groups": [
       "CSS Types"
     ],
-    "status": "standard"
+    "status": "standard",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/angle"
   },
   "basic-shape": {
     "groups": [
       "CSS Shapes",
       "CSS Types"
     ],
-    "status": "standard"
+    "status": "standard",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/basic-shape"
   },
   "blend-mode": {
     "groups": [
       "Compositing and Blending",
       "CSS Types"
     ],
-    "status": "standard"
+    "status": "standard",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/blend-mode"
   },
   "color": {
     "groups": [
       "CSS Color",
       "CSS Types"
     ],
-    "status": "standard"
+    "status": "standard",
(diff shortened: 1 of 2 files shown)


css/types.json
```diff
@@ -9,28 +9,32 @@
...
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/blend-mode"
   },
   "color": {
     "groups": [
       "CSS Color",
       "CSS Types"
     ],
-    "status": "standard"
+    "status": "standard",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color"
```
wbamberg: This should be https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

css/types.json
```diff
@@ -40,33 +44,38 @@
...
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/filter-function"
   },
   "flex": {
     "groups": [
       "CSS Grid Layout",
       "CSS Types"
     ],
-    "status": "standard"
+    "status": "standard",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/flex"
```
wbamberg: -> https://developer.mozilla.org/en-US/docs/Web/CSS/flex_value

css/types.json
```diff
@@ -79,87 +88,101 @@
...
+    "status": "standard",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/percentage"
   },
   "position": {
     "groups": [
       "CSS Types"
     ],
-    "status": "standard"
+    "status": "standard",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/position"
```
wbamberg: -> https://developer.mozilla.org/en-US/docs/Web/CSS/position_value

wbamberg:
> an-plus-b (not sure what this is referring to)

Nor am I :(. In general this seems like a helpful page here: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Types)

ddbeck:
OK, I fixed those clashing links. I did a spot check, but I guess I was too confident. Also, decided to leave off the link for `an-plus-b`. Thanks for the review, @wbamberg!

PR #265, opened 2018-08-18
 Add Japanese translations for column-gap property 

mfuji09:
Add Japanese translations for column-gap property


l10n/css.json
```diff
@@ -1040,6 +1045,7 @@
   },
   "perGrammar": {
     "de": "nach Grammatik",
+    "ja": "構文通り",
```
wbamberg: Should we keep alphabetical order?

mfuji09: Thank you. I fixed it.

PR #269, opened 2018-08-24
image-orientation: make <angle> and flip obsolete

estelle:
dropped support for <angle> and flip

https://bugzilla.mozilla.org/show_bug.cgi?id=1473450

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -4989,7 +4989,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/hyphens"
   },
   "image-orientation": {
-    "syntax": "from-image | <angle> | [ <angle>? flip ]",
+    "syntax": "none | from-image",
     "media": "visual",
     "inherited": true,
     "animationType": "discrete",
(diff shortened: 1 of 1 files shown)


wbamberg:
Is there an updated specification that reflects these changes? The ["specifications" section in the MDN page for `image-orientation`](https://developer.mozilla.org/en-US/docs/Web/CSS/image-orientation#Specifications) links to 2 specs:
* [CSS Images Module Level 3](https://drafts.csswg.org/css-images-3/#the-image-orientation), which has the same syntax as the current version,
* [CSS Images Module Level 4](https://drafts.csswg.org/css-images-4/#image-orientation), which doesn't list this property at all.

The Level 3 spec does say that `<angle>` and `flip` are deprecated, but it does still list them in the syntax. So I'm not sure what the right thing to do here is. I understand that this change better reflects how this property should be used, but I'd be more comfortable if there was an updated spec syntax we could point to.

estelle:
Sorry. From now on I'll add a link to the ticket in my PRs. 

https://bugzilla.mozilla.org/show_bug.cgi?id=1473450

"The <angle> values have been deprecated in the spec, and we are the only browser to implement them anyway."

https://drafts.csswg.org/css-images-3/#the-image-orientation

" This property is likely going to be deprecated and its functionality moved to HTML. At minimum, it will likely lose all but its initial value and from-image."

wbamberg:
Thanks @estelle . You did link to the ticket, and I did see the note you quoted, as well as the note in the spec about deprecating these properties.

But although it deprecates them, the spec still lists `<angle>` and `flip` in the syntax. So I'm not sure if removing them from the syntax in mdn/data is the right thing to do at this point. If there was an updated spec that removed them I'd be happier making this change.

estelle:
Just the note "This property is likely going to be deprecated and its functionality moved to HTML. At minimum, it will likely lose all but its initial value and from-image."
I'll close.

Issue #262, opened 2018-08-16
Expose short descriptions for CSS properties in mdn/data


wbamberg:
This is to address https://github.com/mdn/data/issues/199 for CSS properties.

The overall proposal for how we are going to do this is captured in the [CSS property short descriptions](https://github.com/mdn/data/wiki/CSS-property-short-descriptions) Wiki document.

Broadly there are two aspects to the work: (1) fixing up the content and then (2) getting it into mdn/data:

### Create and test sample short descriptions

(Issue https://github.com/mdn/data/issues/261)

Write short descriptions for a small number (10) of CSS properties, and test how they would work in the different contexts they might appear in (an MDN page, an editor popup, a devtools widget).

On the basis of these samples:

1) draft some guidelines about how the short descriptions should be written
2) refine the estimates of how long it will take to update all of them.

### Update short descriptions for all CSS properties

Go through all the CSS property pages, checking that the summary is what we want: meaning, it's explicitly marked using "seoSummary", and contains the right content.

## Write a script to expose short descriptions in mdn/data

Write a script that works something  like this:

* For each entry in the properties.json file in mdn/data:

    a) Get the corresponding MDN page as a JSON object using the $json parameter (e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/margin$json)

    b) From this object, get the summary property, sanitize it to ensure it only contains the allowed elements and check that it's the correct length.

    c) If the result is different from the current value of shortDescription in properties.json, then add it to a changelist.

 * Open a PR against mdn/data containing the changelist.

PR #276, opened 2018-08-31
Add shortDescription to CSS property schema

ddbeck:
This adds the `shortDescription` as an optional property in the CSS properties schema and documents it too. A small step toward issue #262.


ddbeck:
This is going to be part of [mdn/short-descriptions](https://github.com/mdn/short-descriptions) instead, so closing this up.

PR #277, opened 2018-09-01
Update computed value of overflow* properties

mildsunrise:
Amends #202 and #207 to fix computed value, so it matches [the spec](https://drafts.csswg.org/css-overflow-3/#property-index).

Basically, warn developers that setting `overflow-x: hidden` also means `overflow-y: auto` implicitly.

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6378,7 +6378,10 @@
     ],
     "initial": "visible",
     "appliesto": "blockContainersFlexContainersGridContainers",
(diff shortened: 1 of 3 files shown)


jmendeth:
Hi, could you review this? It'd be nice to fix these errors.

chrisdavidmills:
It's been a long time, but this is still relevant. Thanks @mildsunrise !

PR #279, opened 2018-09-03
feat(api): Add HTMLMarqueeElement inheritance data

ExE-Boss:
This adds inheritance data for [`HTMLMarqueeElement`](https://developer.mozilla.org/docs/Web/API/HTMLMarqueeElement).

---

See also: https://github.com/mdn/browser-compat-data/pull/2714

Also, I’m fine with #240.

### DIFF
--- a/api/inheritance.json
+++ b/api/inheritance.json
@@ -442,6 +442,10 @@
       "SVGUnitTypes"
     ]
   },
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
This looks correct for the current state of things. Thanks!

PR #280, opened 2018-09-04
Update animation type of inset properties

rachelandrew:
Emilio pointed out that the animation type for these properties had changed. Updated to match the spec.


chrisdavidmills:
LGTM; thanks!

PR #283, opened 2018-09-06
feat(css): Add the viewport‑fit @viewport descriptor

ExE-Boss:
This adds data for [the `viewport‑fit` `@viewport` CSS descriptor](https://developer.mozilla.org/docs/Web/CSS/@viewport/viewport-fit).

---

See also mdn/kumascript#802 and mdn/browser-compat-data#2173.

review?(@chrisdavidmills, @ddbeck)

### DIFF
--- a/css/at-rules.json
+++ b/css/at-rules.json
@@ -455,6 +455,18 @@
         "order": "uniqueOrder",
         "status": "standard"
       },
+      "viewport-fit": {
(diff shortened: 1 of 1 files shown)


wbamberg:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1491509

PR #285, opened 2018-09-07
feat(css): Add the `||` column combinator

ExE-Boss:
This adds the syntax data for [the `||` column combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Column_combinator).

review?(@chrisdavidmills, @Elchi3)


chrisdavidmills:
Sounds OK to me. I don't think this is supported anywhere yet, right?

PR #286, opened 2018-09-07
feat(css): Add the `:has(…)` selector

ExE-Boss:
Needed for [the `:has(…)` selector](https://developer.mozilla.org/docs/Web/CSS/:has)

Partially depends on #287, see also https://github.com/mdn/browser-compat-data/pull/1461.

review?(@chrisdavidmills)


chrisdavidmills:
Yup, this LGTM. R+

PR #290, opened 2018-09-10
updating values for display-inside

rachelandrew:
subgrid is no longer a value of display as the subgrid spec has changed.


chrisdavidmills:
LGTM, r+

Thanks @rachelandrew !

PR #293, opened 2018-09-13
<flex> is invalid as a minimum in minmax()

zzzzBov:
[Per the CSS Grid spec](https://www.w3.org/TR/css-grid-1/#valdef-grid-template-columns-minmax)

> minmax(min, max)
> Defines a size range greater than or equal to min and less than or equal to max. If max < min, then max is ignored and minmax(min,max) is treated as min. As a maximum, a <flex> value sets the track’s flex factor; it is invalid as a minimum.


wbamberg:
Thanks for your contribution @zzzzBov ! We're in the middle of trying to relicense this repo to CC0, so if you want your PR to be merged, please visit https://github.com/mdn/data/issues/240 and leave a comment that you agree to the licensing change.

zzzzBov:
@wbamberg all set :+1:

PixleProgrambetaxl:
+1

chrisdavidmills:
It's been a long time, but this is still relevant; thanks @zzzzBov !

PR #294, opened 2018-09-13
removing certain contributions as part of relicensing plan

chrisdavidmills:
Here we are removing certain contributions to allow us to complete our plan to relicense the repo to CC0 (see https://github.com/mdn/data/issues/240 for more context).

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5315,22 +5315,6 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/line-height"
   },
-  "line-height-step": {
-    "syntax": "<length>",
-    "media": "visual",
-    "inherited": true,
-    "animationType": "discrete",
-    "percentages": "no",
-    "groups": [
-      "CSS Fonts"
-    ],
-    "initial": "0",
-    "appliesto": "blockContainerElements",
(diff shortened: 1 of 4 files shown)


chrisdavidmills:
So I removed angle, but I think AsAngleBasicshapeOrPath was already updated to angleOrBasicShapeOrPath. I therefore removed that one instead.

chrisdavidmills:
I've looked for dependencies on those...I think that's about it?

wbamberg:
It should also be removed from the schema: https://github.com/mdn/data/search?q=angleOrBasicShapeOrPath. Sorry!

wbamberg:
>  I believe offset-path is already removed, so the instance in the properties.json file is not there?

Yes, I think that's already removed in this PR: https://github.com/mdn/data/pull/294/files#diff-55f6403f7fb454999a3f946361507bdeL6164 although the diffs are confusing.

octref:
Can I ask why brackets are escaped here?
Also, can these removed properties be brought back?

chrisdavidmills:
@octref I don't really know why brackets were escaped. That is a bit strange.

I have added back a few of the removed properties already, in https://github.com/mdn/data/pull/309. Happy to add others back.

PR #297, opened 2018-09-14
Add script to check MDN page summary/short description

ddbeck:
This PR adds a script that lets you check a CSS property's MDN page summary, following that property's `mdn_url` value.

Things to highlight while reviewing:
* I'm pretty new to JavaScript and Node.js. I'd welcome any comments on style, idiom, etc.
* I added two dependencies. I added `jsdom` to get access to `textContent` and to make finding tags and attributes a bit easier. And I added `request` to make fetching the MDN page a little easier—though I think this one might be possible to avoid with Node's `https` module (at the expense of a longer script).
* Once we get actual short descriptions into the data itself, we'll probably want to switch to checking the shortDescription field instead of fetching the MDN page (and use a different script to fetch the pages and update shortDescription). I think this is a first iteration, not necessarily the final say on how to check these things.

Thanks for your patience on this, @wbamberg. I had to learn more new things than I expected and I had less time than I expected.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,27 @@
         "js-tokens": "^3.0.0"
       }
     },
+    "abab": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-2.0.0.tgz",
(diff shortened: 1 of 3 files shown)


ddbeck:
Oh yeah one more thing: I agreed to #240 already, but to reiterate, I'm happy with this being made available under either the old (MPL) or new license (CC0).

ddbeck:
This work is being relocated to https://github.com/mdn/short-descriptions/pull/4, so I'm closing this up.

Issue #292, opened 2018-09-12
`place-self` property is missing


davidbernegger:
Just like #228, property is missing in css/properties.json

frenic:
Duplicates #253

davidbernegger:
Thanks, thought I searched for it, gonna close the issue! 👍

Issue #253, opened 2018-08-07
`place-self` property missing


connorshea:
This property isn't included in the data repository despite being standard: https://developer.mozilla.org/en-US/docs/Web/CSS/place-self

chrisdavidmills:
Fixed by https://github.com/mdn/data/pull/304

octref:
Hey @chrisdavidmills, do you mind publishing a new version so I can fix https://github.com/Microsoft/vscode/issues/64835? Thanks in advance.

wbamberg:
@octref , @chrisdavidmills , this should be done now.

octref:
Adopted in latest version. Thanks a lot!

chrisdavidmills:
@wbamberg yay, cheers Will!

PR #304, opened 2018-10-02
Adding place-self property

davidbernegger:
As requested #253 #292 and  Microsoft/vscode-css-languageservice#117

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6819,7 +6819,7 @@
       "align-self",
       "justify-self"
     ],
-    "appliesto": "allElements",
+    "appliesto": "blockLevelBoxesAndAbsolutelyPositionedBoxesAndGridItems",
     "computed": [
       "align-self",
       "justify-self"
--- a/css/properties.json
+++ b/css/properties.json
@@ -6806,6 +6806,28 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/place-items"
   },
+  "place-self": {
+    "syntax": "<'align-self'> <'justify-self'>?",
+    "media": "visual",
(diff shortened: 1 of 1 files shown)


connorshea:
For reviewers:

- https://drafts.csswg.org/css-align-3/#place-self-property
- https://developer.mozilla.org/en-US/docs/Web/CSS/place-self

css/properties.json
```diff
@@ -6806,6 +6806,28 @@
...
+    "animationType": "discrete",
+    "percentages": "no",
+    "groups": [
+      "CSS Box Alignment"
+    ],
+    "initial": [
+      "align-self",
+      "justify-self"
+    ],
+    "appliesto": "allElements",
```
connorshea: The spec says that it applies to "block-level boxes, absolutely-positioned boxes, and grid items", not all elements.

https://www.w3.org/TR/css-align-3/#place-self-property

octref:
Just subscribe myself here, also for https://github.com/Microsoft/vscode/issues/64835.

css/properties.json
ExE-Boss:
```suggestion
    "appliesto": "blockLevelBoxesAndAbsolutelyPositionedBoxesAndGridItems",
```

ExE-Boss:
You can just accept my suggestion from the&nbsp;Github&nbsp;UI.

davidbernegger:
Just updated it with your suggestion @ExE-Boss

chrisdavidmills:
This looks good; thanks @davidbernegger !

davidbernegger:
@chrisdavidmills:  @connorshea's pull request #253 could be closed.

PR #311, opened 2018-10-16
addition of clamp(), min(), and max()

estelle:
https://www.w3.org/TR/css-values/#calc-syntax


chrisdavidmills:
I've checked this against the spec (https://www.w3.org/TR/css-values/#calc-syntax), and it looks correct to me. r+; cheers @estelle !

PR #312, opened 2018-10-17
line-break is an inherited property

eiriarte:
As per the linked spec: https://drafts.csswg.org/css-text-3/#line-break-property


chrisdavidmills:
@eiriarte only just saw this one; looks good, thanks!

PR #313, opened 2018-10-18
Add line-height-step data

ddbeck:
While using the data to look up URLs, I found that this property was missing entirely. This is my first time adding a property from scratch, so I'd appreciate a closer look here. Thank you!

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5329,7 +5329,7 @@
     "computed": "absoluteLength",
     "order": "perGrammar",
     "status": "experimental",
-    "mdn_url": "https://developer.mozilla.org/docs/Mozilla/CSS/line-height-step"
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/line-height-step"
   },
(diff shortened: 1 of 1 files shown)


css/properties.json
```diff
@@ -5315,6 +5315,22 @@
...
+    "percentages": "no",
+    "groups": [
+      "CSS Fonts"
+    ],
+    "initial": "0",
+    "appliesto": "blockContainers",
+    "computed": "absoluteLength",
+    "order": "perGrammar",
+    "status": "experimental",
+    "mdn_url": "https://developer.mozilla.org/docs/Mozilla/CSS/line-height-step"
```
frenic: I got a 404 with this. The correct one seems to be: https://developer.mozilla.org/docs/Web/CSS/line-height-step

ddbeck: Good catch, thank you! Fixed!

chrisdavidmills:
Reviewed, and looks fine to me.

PR #317, opened 2018-10-23
Drop motion-* properties. Fix #314

### LINKED ISSUE #314: Drop motion in favor of offset
From this page https://developer.mozilla.org/en-US/docs/Web/CSS/offset and the spec, `motion` is changed to `offset` in the spec proposal.

Should I remove the `motion` properties, or should I add a deprecated status for them? (I think it makes more sense to remove them, since these properties are dropped in the experimental phase).

octref:


pyoor:
Looks like I missed this when I push https://github.com/mdn/data/pull/349.  Now that that has landed, this can be closed.

chrisdavidmills:
OK, closing.

PR #319, opened 2018-10-30
Adding inset property

rachelandrew:
The inset property is missing from the data, I think this may be the cause of the inset() basic shape being used incorrectly here: https://developer.mozilla.org/en-US/docs/Web/CSS/inset#Formal_syntax

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5100,7 +5100,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/inline-size"
   },
   "inset": {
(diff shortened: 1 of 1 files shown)


rachelandrew:
which longhand properties though? The t,r,b,l ones or the logical versions. The spec just points to the physical ones.

rachelandrew:
Also, it looks like the other ones are using left not top, so should I change those as well?

chrisdavidmills:
LGTM. Thanks Rachel.

PR #321, opened 2018-11-05
Scroll snap data

rachelandrew:
Adding the scroll-snap properties which are missing https://drafts.csswg.org/css-scroll-snap-1/

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -7023,7 +7023,7 @@
     "animationType": "byComputedValueType",
     "percentages": "no",
     "groups": [
-      "CSS ScrollSnap"
+      "CSS Scroll Snap"
     ],
     "initial": "0",
     "appliesto": "allElements",
@@ -7039,7 +7039,7 @@
     "animationType": "byComputedValueType",
     "percentages": "no",
     "groups": [
-      "CSS ScrollSnap"
+      "CSS Scroll Snap"
     ],
     "initial": "0",
     "appliesto": "allElements",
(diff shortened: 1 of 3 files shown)


rachelandrew:
This is failing due to ENUM data which going by the spec data doesn't seem to have a match. Animation on some properties is specified as "by computed value type" https://drafts.csswg.org/css-scroll-snap-1/#scroll-padding I don't think is is included. Can I add it or is it represented by another type?

rachelandrew:
I have the same issue with specifying the percentages.

chrisdavidmills:
So for percentages, there shouldn't be a problem..."no" is already specified at https://github.com/mdn/data/blob/master/css/properties.schema.json#L49.

for the animation type, I don't think we have this one already; feel free to add it here:

https://github.com/mdn/data/blob/master/css/properties.schema.json#L16

You then also need to add an entry for it in https://github.com/mdn/data/blob/master/l10n/css.json

rachelandrew:
there is also the percentage values for scroll-padding. Those are the ones that are failing. 'no' is fine

https://www.w3.org/TR/css-scroll-snap-1/#padding-longhands-physical

chrisdavidmills:
> there is also the percentage values for scroll-padding. Those are the ones that are failing.

Ooops, missed that one, sorry!

So yeah, same story. Feel free to add it into 

* https://github.com/mdn/data/blob/master/css/properties.schema.json#L40
* https://github.com/mdn/data/blob/master/l10n/css.json

rachelandrew:
ok - this is passing now

chrisdavidmills:
Perfect, thanks!

PR #323, opened 2018-11-07
Moving box-align properties out of the flexbox group

rachelandrew:
The flexbox spec explains that we should refer to Box Align in future for these properties and we have now updated the docs to explain the relationship. So moving the properties from flexbox to box align in the data.


chrisdavidmills:
LGTM

PR #287, opened 2018-09-07
feat(css): Add <complex-selector> syntax

ExE-Boss:
See also #286 and #284.

`<compound-selector>` was already added in PR&nbsp;https://github.com/mdn/data/pull/169.

Blocks https://github.com/mdn/data/pull/324.

review?(@chrisdavidmills, @Elchi3)

PR #324, opened 2018-11-15
renaming :matches to :is

chrisdavidmills:
As per https://github.com/w3c/csswg-drafts/issues/3258

### DIFF
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -342,14 +342,14 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:link"
   },
-  ":matches": {
-    "syntax": ":matches( <selector># )",
+  ":is": {
+    "syntax": ":is( <selector># )",
     "groups": [
       "Pseudo-classes",
       "Selectors"
     ],
     "status": "standard",
-    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:matches"
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:is"
   },
   ":not": {
     "syntax": ":not( <selector># )",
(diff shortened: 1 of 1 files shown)


css/selectors.json
```diff
@@ -342,14 +342,14 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:link"
   },
-  ":matches": {
-    "syntax": ":matches( <selector># )",
+  ":is": {
+    "syntax": ":is( <selector># )",
```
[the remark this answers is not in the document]
chrisdavidmills: Sure; updated.

css/selectors.json
```diff
@@ -342,14 +342,14 @@
...
   },
-  ":matches": {
-    "syntax": ":matches( <selector># )",
+  ":is": {
+    "syntax": ":is( <complex-selector># )",
     "groups": [
       "Pseudo-classes",
       "Selectors"
     ],
     "status": "standard",
```
ExE-Boss: Also, this is probably `experimental` now.

chrisdavidmills: Ack, forgot about that, thanks! I did set it as experimental in BCD.

css/selectors.json
```diff
@@ -342,14 +342,14 @@
...
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:link"
   },
-  ":matches": {
-    "syntax": ":matches( <selector># )",
+  ":is": {
+    "syntax": ":is( <complex-selector># )",
```
ExE-Boss: I’m sort to bug you a third time, but I realised that this should actually be `<complex-selector-list>`, like in https://github.com/mdn/data/pull/287: https://github.com/mdn/data/blob/83365d0f7c8665c2fb143d157bf0dc47ac598cec/css/selectors.json#L354

chrisdavidmills:
Ah, so that's the official syntax that represents a complex selector list, so we don't need to write "one or more complex selectors". Works for me. Updated.

chrisdavidmills:
Done and merged. Thanks!

PR #326, opened 2018-11-22
feat(css): Add the `env(…)` function syntax

ExE-Boss:
review?(@chrisdavidmills)

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -197,6 +197,9 @@
   "ending-shape": {
     "syntax": "circle | ellipse"
   },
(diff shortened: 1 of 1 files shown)


ExE-Boss:
@chrisdavidmills But that would require `prop: env(safe-area-inset-top,)` to be valid and `prop: env(safe-area-inset-top)` to be invalid, but as we all know, it’s the other way around, so the spec is incorrect.

chrisdavidmills:
Thanks @tabatkins for the clarification. This one had me confused.

PR #329, opened 2018-11-28
Adding two value shorthands

rachelandrew:
The two value shorthands were missing for Logical Properties, these don't currently have any browser support.

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -2261,6 +2261,78 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/border"
   },
+  "border-block": {
+    "syntax": "<'border-width'> || <'border-style'> || <'color'>",
+    "media": "visual",
+    "inherited": false,
+    "animationType": "discrete",
+    "percentages": "no",
+    "groups": [
+      "CSS Logical Properties"
+    ],
+    "initial": [
+      "border-top-width",
+      "border-top-style",
+      "border-top-color"
+    ],
+    "appliesto": "allElements",
+    "computed": [
+      "border-top-width",
+      "border-top-style",
+      "border-top-color"
+    ],
+    "order": "uniqueOrder",
+    "status": "standard",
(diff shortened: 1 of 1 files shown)


css/properties.json
```diff
@@ -2261,6 +2261,78 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/border"
   },
+  "border-block": {
+    "syntax": "<'border-width'> || <'border-style'> || <'color'>",
```
chrisdavidmills: according to the spec, should be `<'border-top-width'> || <'border-top-style'> || <'color'>` ?

And same pattern for the other properties that have the same syntax?

css/properties.json
```diff
@@ -2748,6 +2820,30 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/border-image-width"
   },
+  "border-inline": {
+    "syntax": "<'border-width'> || <'border-style'> || <'color'>",
```
chrisdavidmills: according to the spec, should be `<'border-top-width'> || <'border-top-style'> || <'color'>` ?

css/properties.json
```diff
@@ -2261,6 +2261,78 @@
...
+    "computed": [
+      "border-top-width",
+      "border-top-style",
+      "border-top-color"
+    ],
+    "order": "uniqueOrder",
+    "status": "standard",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/border-block"
+  },
+  "border-block-color": {
```
chrisdavidmills: Would it make more sense for the syntax to be `<'color'>{1,2}`, as you can specify one or two colors — both the same or separate colors for each, can't you? And same pattern for `*-style` and `*-width`?

Also same pattern for the inline versions?

rachelandrew: so it looks like the spec isn't consistent with these. These ones use border-top-color, the border-block border-inline properties just use color. I've updated everything to match the spec even if that seems inconsistent.

PR #331, opened 2018-11-29
Fix -webkit-overflow-scrolling inheritance

timothy003:
This corrects -webkit-overflow-scrolling to be an inherited property. Setting -webkit-overflow-scrolling: touch on an element causes it and its descendants to create stacking contexts, which affects the behavior of z-index.

https://bugs.chromium.org/p/chromium/issues/detail?id=128325#c23

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -1505,7 +1505,7 @@
   "-webkit-overflow-scrolling": {
     "syntax": "auto | touch",
     "media": "visual",
-    "inherited": false,
+    "inherited": true,
     "animationType": "discrete",
     "percentages": "no",
     "groups": [


timothy003:
Should `appliesto` also be changed to `"allElements"` to reflect its behavior on non-scrolling boxes?

chrisdavidmills:
So, I agree with you about inheritance — it sounds like at least some significant behavior is inherited.

As for "appliesto", I think it'd make more sense to leave it as is. It doesn't really make sense to apply this to a non-scrolling element? What behavior does it exhibit when applied directly to a non-scrolling element?

timothy003:
According to [this discussion](https://bugs.webkit.org/show_bug.cgi?id=103598), it used to create a stacking context even for non-scrolling elements, but after testing the fiddle on iOS 12.1, this no longer appears to be the case. So, I assume that's been fixed and it now only creates a stacking context for scrollable elements.

chrisdavidmills:
OK, cool. I think this is good as is then...merging.

Thanks for the contribution!

PR #333, opened 2018-12-20
Fix border-bottom-color reference syntax

frenic:
I got an error when building types with CSSType using your latest changes. I understand this is the way you reference the syntax to another property.


chrisdavidmills:
Ah, good catch — thanks.

PR #336, opened 2018-12-28
Conicgradient

estelle:
added conic gradient to the gradient list, then added all the new synstax features that make up the conic gradient.

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -8,6 +8,12 @@
   "angle-percentage": {
     "syntax": "<angle> | <percentage>"
   },
+  "angular-color-stop": {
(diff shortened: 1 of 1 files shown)


estelle:
that's so weird. I somehow reverted everything from 
9db25b8  with f280fca.

Crying is much easier than git, but isn't as effective I hear. ugh.

estelle:
Updated. Great catch.

chrisdavidmills:
Nearly there; still need to update `<color-stop>` syntax to `<color-stop-length> | <color-stop-angle>`

estelle:
Updated. Thanks. BTW, I don't think <color-stop> is used anywhere anymore.

PR #340, opened 2019-01-07
fixing env() formal syntax

chrisdavidmills:
As per https://github.com/mdn/data/pull/326#issuecomment-451526981


ExE-Boss:
This will also have to&nbsp;be&nbsp;done for&nbsp;`var(…)` once&nbsp;https://github.com/w3c/csswg-drafts/issues/3485 is&nbsp;merged.

PR #345, opened 2019-01-15
Updating break-inside, break-before and break-after

rachelandrew:
Media to visual as these apply not just in paged contexts. Added the always and all values from the Level 4 spec. https://github.com/w3c/csswg-drafts/issues/3318


chrisdavidmills:
Looks fine; thanks!

PR #346, opened 2019-01-16
added step jump values. Removed the 'single' from <timing-function>

estelle:
add step- https://bugzilla.mozilla.org/show_bug.cgi?id=1496619
syntax at https://drafts.csswg.org/css-easing-1/#typedef-step-timing-function

https://drafts.csswg.org/css-animations/#typedef-single-animation - change single-timing-functon to timing-function for animation property.

https://drafts.csswg.org/css-animations/#animation-timing-function - change single-timing-functon to timing-function for animation-timing-function property.

https://drafts.csswg.org/css-transitions/#transition-timing-function-property - change single-timing-functon to timing-function for transition-timing-function property.

https://drafts.csswg.org/css-transitions/#single-transition - do the same for transition.

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -1849,7 +1849,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/animation-play-state"
   },
   "animation-timing-function": {
-    "syntax": "<single-timing-function>#",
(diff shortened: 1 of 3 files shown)


chrisdavidmills:
Looked good to me

PR #347, opened 2019-01-20
syntax order to match spec

estelle:
https://drafts.csswg.org/css-transforms-2/#propdef-rotate

the old syntax and new are actually the same in the end, but the update is easier to decipher and matches what is currently in the spec linked to above.


chrisdavidmills:
Looks fine as per spec.

PR #43, opened 2017-02-01
Add offset properties to CSS.

jpmedley:
The replace motion properties. So: 

motion-path becomes offset-path
motion-offset becomes offset-distance
motion-rotation becomes offset-rotation
motion becomes offset

The old ones aren't removed because I don't know if they're used anywhere.

PR #349, opened 2019-01-28
Remove invalid motion-* properties

pyoor:
The motion-* properties were initially removed via PR #43 but then re-added in https://github.com/mdn/data/commit/582f2567ceafa36437e0b32f680f16cf5b6285f5.  This appears to be accidental but if not, the values must be stripped of HTML entities.


chrisdavidmills:
This looks good, thanks!

PR #354, opened 2019-02-26
Which elements overflow-wrap applies to.

estelle:
Currently reads 'all elements' but spec says 'inline boxes'.
https://drafts.csswg.org/css-text-3/#propdef-overflow-wrap
An inline box is a non-replaced inline-level box whose inner display type is flow. 
for #937


chrisdavidmills:
WFM.

PR #362, opened 2019-04-03
test: Use `ora` to make lint output less verbose

ExE-Boss:
This copies the changes from https://github.com/mdn/browser-compat-data/pull/2528.

(Preferably, we should keep common functions like schema and style validation for MDN repos in a scoped `@mdn/*` package)


.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: node_js
 node_js: stable
 sudo: false
+before_install: "curl -L https://unpkg.com/@pnpm/self-installer | node"
```
lahmatiy: What's the reason to use `pnpm`? I think, it makes things more complicated where unnecessary

test/lint.js
```diff
@@ -61,43 +100,77 @@ function checkSchema(dataFilename) {
...
-      // Other issues with better-ajv-errors:
-      // - it feels better for performance to output messages one by one rather than a list
-      // - it seems to be losing some errors when output a list
-      ajv.errors.forEach(function(error) {
-        var message = betterAjvErrors(schema, data, [error], { indent: 2 });
-
-        console.log('\n    ' + message.replace(/\n/g, '\n    '));
+      console.error(`\x1b[31m  JSON Schema – ${ajv.errors.length} error(s)\x1b[0m`)
+      ajv.errors.forEach(e => {
+        // Output messages by one since better-ajv-errors wrongly joins messages
```
lahmatiy: This comment is for forEach not for the function body

test/lint.js
```diff
@@ -33,26 +66,32 @@ function jsonDiff(actual, expected) {
...
   var actual = fs.readFileSync(filename, 'utf-8').trim();
   var expected = JSON.stringify(JSON.parse(actual), null, 2);
 
   if (actual === expected) {
-    console.log('  Style – OK');
+    return false;
   } else {
-    hasErrors = true;
-    console.log('  Style – Error on line ' + jsonDiff(actual, expected));
+    console.error(`\x1b[31m  Style – Error on line ${jsonDiff(actual, expected)}\x1b[0m`);
```
lahmatiy: It's better to use something like `chalk` package to highlight output

test/lint.js
```diff
@@ -61,43 +100,77 @@ function checkSchema(dataFilename) {
...
-      checkStyle(absFilename)
-      checkSchema(absFilename);
+      const spinner = ora({
+        stream: process.stdout,
+        text: relativeFilePath
+      });
 
-      console.log();
+      const console_error = console.error;
+      console.error = (...args) => {
```
lahmatiy: That's a bad idea to monkey patching a console method. `testStyle` and `testSchema` could return an array of errors if any instead of calling `console.log()`. Later we can concat these arrays and if a resulting array is not empty then there is a file with errors.

test/lint.js
```diff
@@ -61,43 +100,77 @@ function checkSchema(dataFilename) {
...
   });
 });
 
 if (hasErrors) {
+  console.warn("");
+  console.warn(`Problems in ${filesWithErrors.size} file${filesWithErrors.size > 1 ? 's' : ''}:`);
+  for (const [fileName, file] of filesWithErrors) {
+    console.warn(fileName);
+    try {
+      testSchema(file);
```
lahmatiy: Looks like a double checking of files and message output.

.travis.yml
ExE-Boss: [pnpm](https://pnpm.js.org) is the third largest node package manager, which uses symlinks and isolated trees to&nbsp;prevent code from accessing packages which they haven’t declared in their dependencies.

**See also:**
- https://medium.com/pnpm/pnpms-strictness-helps-to-avoid-silly-bugs-9a15fb306308
- https://www.kochan.io/nodejs/why-should-we-use-pnpm.html

test/lint.js
```diff
@@ -25,6 +31,33 @@ ajv.addKeyword('property-reference', {
 
 ajv.addSchema(require('../css/definitions.json'), 'definitions.json');
 
+/**
+ * @param {string} str
+ */
+function escapeInvisibles(str) {
+  const invisibles = [
```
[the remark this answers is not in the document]
ExE-Boss: This was copied from:
https://github.com/mdn/browser-compat-data/blob/9d92e6d7f88cfa4400423a39183c321d7a68c410/test/test-style.js#L28-L45

test/lint.js
ExE-Boss: Take this up&nbsp;with the&nbsp;maintainers of&nbsp;https://github.com/mdn/browser-compat-data/blob/9d92e6d7f88cfa4400423a39183c321d7a68c410/test/lint.js#L52-L58

test/lint.js
ExE-Boss: Again, this is from https://github.com/mdn/browser-compat-data/blob/9d92e6d7f88cfa4400423a39183c321d7a68c410/test/lint.js#L113-L132.

.travis.yml
vinyldarkscratch: Linking the related conversation in mdn/mdn here: https://github.com/mdn/mdn/pull/69

PR #364, opened 2019-04-16
fixing attr() formal syntax

chrisdavidmills:
As per https://github.com/mdn/sprints/issues/1155

I made this fix, but I also added some of the missing type-or-unit options that had been added since.

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -735,7 +735,7 @@
     "syntax": "translateZ( <length> )"
   },
   "type-or-unit": {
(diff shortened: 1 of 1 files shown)


estelle:
Looks good other than perhaps the order of where 'px' is placed: not sure which order we wanted to go with for the units: it seems kind of by subgroups like relative sizes, font-related sizes, fixed sizes, time, frequences, angles. If that is hte case, px should be with fixed - cm, in, mm, Q, pc, pt, px.

The information looks fine (is approved), just not sure if you wanted to change the order.

chrisdavidmills:
Good call @estelle ; I've moved it ot a better place. All good now?

chrisdavidmills:
Cool, let's get this merged. Thanks for the review @ExE-Boss

PR #376, opened 2019-06-27
Sync transform function syntaxes with latest changes in specs

lahmatiy:
Changes:
- In `rotate()`, `rotateX()`, `rotateY()`, `rotateZ()`, `skew()`, `skewX()` and `skewY()`: `<angle>` is replaced for `[ <angle> | <zero> ]`.
- In `matrix()` and `matrix3d()`: simplify syntax by using `#` multiplier (see https://github.com/w3c/csswg-drafts/pull/4053, https://github.com/w3c/csswg-drafts/pull/4054)
- In `skew`, `scale()` and `translate()`: remove redundant grouping (brackets) (see https://github.com/w3c/csswg-drafts/pull/4054)

References:
- https://drafts.csswg.org/css-transforms/#two-d-transform-functions
    * `matrix()`
    * `rotate()`
    * `scale()`
    * `skew()`
    * `skewX()`
    * `skewY()`
    * `translate()`
- https://drafts.csswg.org/css-transforms-2/#three-d-transform-functions
    * `matrix3d()`
    * `rotate3d()`
    * `rotateX()`
    * `rotateY()`
    * `rotateZ()`

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -414,10 +414,10 @@
     "syntax": "alpha | luminance | match-source"
   },
   "matrix()": {
-    "syntax": "matrix( <number> [, <number> ]{5,5} )"
+    "syntax": "matrix( <number>#{6} )"
   },
   "matrix3d()": {
-    "syntax": "matrix3d( <number> [, <number> ]{15,15} )"
+    "syntax": "matrix3d( <number>#{16} )"
   },
   "max()": {
     "syntax": "max( <calc-sum># )"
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
looks good also, thanks!

lahmatiy:
@chrisdavidmills Could you publish a new version please? There are many important fixes accumulated. For example, https://github.com/csstree/csstree/issues/99#issuecomment-509207350

squidfunk:
@lahmatiy do you know why `<zero>` was added as an explicit type? To my knowledge the unit of a dimension can be omitted if it evaluates to literal zero (`"0"`), so the `<zero>` should be implicit. Also, this is not reflected in other parts of the syntax, e.g. `<length-percentage>` which may be used in `margin` or other properties and can also be zero. I understand that this repository tries to adhere close to the docs, but as it's currently only in ED, I question the value of adding `<zero>` to a subset of all effectively necessary syntaxes.

lahmatiy:
@squidfunk Using zero as a unitless value is [allowed](https://drafts.csswg.org/css-values-3/#lengths) for `<length>` only (it covers any `<length>` production, e.g. `<length-percentage>`):

> For zero lengths the unit identifier is optional (i.e. can be syntactically represented as the `<number>` `0`).

At one time a unitless zero was valid as `<angle>` value in some syntaxes. However, currently unitless zero as `<angle>` is invalid in general, but valid in some syntaxes due to legacy reasons ([spec](https://drafts.csswg.org/css-values-3/#angles)):

> Note: For legacy reasons, some uses of `<angle>` allow a bare `0` to mean `0deg`. This is not true in general, however, and will not occur in future uses of the `<angle>` type.

`<zero>` introduced to indicate explicitly where it allowed. It [appears](https://github.com/w3c/csswg-drafts/commit/602789171429b2231223ab1e5acf8f7f11652eb3#diff-c97a034bff454c2c7b7f7cfe7d2b8d67) in [Values & Units L4](https://drafts.csswg.org/css-values-4/#numbers) (it was marked as `revert`, I didn't found anything about a decision):

> The value `<zero>` represents a literal number with the value `0`. Expressions that merely evaluate to a `<number>` with the value `0` (for example, `calc(0)`) do not match `<zero>`; only literal `<number-token>`s do.

Btw, unitless zero `<length>` is [prohibited in math functions](https://drafts.csswg.org/css-values-4/#calc-type-checking) (e.g. `calc()`):

> Because `<number-token>`s are always interpreted as `<number>`s or `<integer>`s, "unitless 0" `<length>`s aren’t supported in math functions. That is, `width: calc(0 + 5px);` is invalid, because it’s trying to add a `<number>` to a `<length>`, even though both `width: 0;` and `width: 5px;` are valid.

and also has [lower priority](https://drafts.csswg.org/css-values-4/#lengths) over `<number>`: 

> if a `0` could be parsed as either a `<number>` or a `<length>` in a property (such as line-height), it must parse as a `<number>`.

You can meet this case in `flex` property, which has following syntax:
image

So for value `flex: 0 0` second component (`0`) matches on `flex-shrink`:
image

But for value `flex: 0 0px` second component (`0px`) matches on `flex-basis`:
image

squidfunk:
@lahmatiy thank you for the very insightful and detailed clarification!

PR #378, opened 2019-07-02
Fixed color-stop syntax used in gradient functions

squidfunk:
The color stop syntax definition is incorrect, as it doesn't allow for colors stops without lengths, i.e.

``` css
div {
  background: radial-gradient(red 0%, blue 100%); /* works */
  background: radial-gradient(red, blue); /* doesn't work */
}
```

The [MDN documentation][1] states the correct syntax. This PR fixes that.

  [1]: https://developer.mozilla.org/en-US/docs/Web/CSS/radial-gradient#Formal_syntax

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -393,7 +393,7 @@
     "syntax": "<length-percentage>"
   },
   "linear-color-stop": {
-    "syntax": "<color> && <color-stop-length>"
+    "syntax": "<color> <color-stop-length>?"
   },
   "linear-gradient()": {
     "syntax": "linear-gradient( [ <angle> | to <side-or-corner> ]? , <color-stop-list> )"


chrisdavidmills:
Makes sense; thanks!

lahmatiy:
It would be nice to make a PR to spec, since it has `<color> && <color-stop-length>` syntax now: https://www.w3.org/TR/css-images-4/#typedef-linear-color-stop

squidfunk:
I did some research:

- [WD 2012](https://www.w3.org/TR/2012/WD-css4-images-20120911/#color-stop-syntax): `<color> [ <percentage> | <length> ]{0,2}` – supported by all major browsers, though I don't know why there are two numbers. Shouldn't it be `?` instead of `{0,2}`?
- [WD 2017](https://www.w3.org/TR/css-images-4/#color-stop-syntax):  `<color> && <color-stop-length>` – clearly wrong, as length is now mandatory and position independent, supported by no browser at all
- [ED](https://drafts.csswg.org/css-images-4/#typedef-linear-color-stop): `<color> && <color-stop-length>?` – makes more sense, but still no browser support.

So the ED corrects the latest WD, but it's still an ED, so I would propose leaving it as originally proposed by this PR.

lahmatiy:
> though I don't know why there are two numbers

`{0,2}` means zero, one or two percentage or length. Double positions [defined in spec](https://www.w3.org/TR/css-images-4/#color-stop-syntax) this way:

> A color stop with two locations is mostly equivalent to specifying two color stops with the same color, one for each position. Specifying two locations makes it easier to create solid-color "stripes" in a gradient, without having to repeat the color twice.

[Browser support](https://developer.mozilla.org/en-US/docs/Web/CSS/linear-gradient#Browser_compatibility) is not complete but solid:
image

Since `&&` is used since 2017 it means that spec allows using color and position in any order now. However, as I know, it doesn't supported by any browser currently.

squidfunk:
Thanks for researching. So I guess we should change the syntax to `<color> <color-stop-length>{0,2}`. Should I create a PR?

lahmatiy:
I found the issue about missed `?` in color stop syntax: https://github.com/w3c/csswg-drafts/issues/1334
However I didn't found when transition `<color> <length-percentage>?` -> `<color> && <color-stop-length>` (`?` were added later) happened. Looks like `&&` appeared from the beginning of CSS Images L4.

> Since `&&` is used since 2017 it means that spec allows using color and position in any order now. However, as I know, it doesn't supported by any browser currently.

@tabatkins @AmeliaBR Could you clarify please, should we create an issue/PR to CSS Images L4 or `<color> && <color-stop-length>?` is a correct syntax for color stop?

PR #384, opened 2019-09-07
fixed link's ref from 'properties.md'

kluid:
While reading 'properties.md', found a links ref is missed.


chrisdavidmills:
Looks reasonable; thanks!

PR #393, opened 2019-10-12
CSS Selectors: add "Grouping Selectors"

ramiy:
Related:

https://github.com/mdn/sprints/issues/2233
https://github.com/mdn/sprints/issues/2164

### DIFF
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -44,14 +44,14 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/Attribute_selectors"
   },
-  "Grouping selectors": {
+  "Selector list": {
     "syntax": ",",
     "groups": [
       "Grouping Selectors",
       "Selectors"
     ],
     "status": "standard",
-    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/Grouping_selectors"
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/Selector_list"
   },
   "Adjacent sibling combinator": {
     "syntax": "A + B",
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -44,6 +44,15 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/Attribute_selectors"
   },
+  "Grouping selectors": {
+    "syntax": ",",
+    "groups": [
+      "Grouping Selectors",
+      "Selectors"
+    ],
+    "status": "standard",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/Grouping_selectors"
+  },
   "Adjacent sibling combinator": {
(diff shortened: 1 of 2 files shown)


chrisdavidmills:
The Travis run failed because you are trying to use a group type that doesn't exist. I'd suggest just using `Basic Selectors` instead of `Grouping Selectors`, as it is a pretty basic type.

ramiy:
@chrisdavidmills 

How can I add a new group type for `Grouping Selectors`.

See related links:

https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors

https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/Universal_selectors
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/Type_selectors

and the new grouping selectors:

https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/Grouping_selectors

css/selectors.json
```diff
@@ -44,6 +44,15 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/Attribute_selectors"
   },
+  "Grouping selectors": {
```
ExE-Boss:
```suggestion
  "Selector list": {
```

css/selectors.json
```diff
@@ -44,6 +44,15 @@
...
+    "syntax": ",",
+    "groups": [
+      "Grouping Selectors",
+      "Selectors"
+    ],
+    "status": "standard",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/Grouping_selectors"
```
ExE-Boss:
```suggestion
    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/Selector_list"
```

chrisdavidmills:
Ah, thanks for sorting that out @ExE-Boss . You are right that we should use Selector list.

@ramiy to add new group type, you add it to the group list at

https://github.com/mdn/data/blob/e82e4caa73d88b0edcb52ae579c75e479c320997/css/definitions.json

css/selectors.json
ramiy: Done.

css/selectors.json
ramiy: Done.

ExE-Boss:
Next, we’ll need [to update KumaScript’s `CSSRef` macro](https://github.com/mdn/kumascript/pull/1256) to account for the new sidebar group, like I did for combinators in mdn/kumascript#942.

ExE-Boss:
A new version of [mdn‑data](https://www.npmjs.com/package/mdn-data) has to be published before this will show up on **MDN** (because of https://github.com/mdn/kumascript/pull/1126)

ramiy:
The last version released [over 3 months ago](https://www.npmjs.com/package/mdn-data?activeTab=versions)... It's time to release a new version.

PR #395, opened 2019-10-19
CSS Selectors - add "Pseudo" group

ramiy:
Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors

In #393 PR we added the Selectors list.

In this PR we we will add a "Pseudo" group linking to:
* [Pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes)
* [Pseudo-elements](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements)

In a separate PR I will move the Pseudo-classes and the Pseudo-elements to `properties.json`. They should not be part of CSS Selectors sidebar.

The final goal is to create a clear hierarchy for CSS Selectors and a separate sidebar for Pseudo-classes and the Pseudo-elements.

### DIFF
--- a/css/definitions.json
+++ b/css/definitions.json
@@ -65,6 +65,7 @@
       "Microsoft Extensions",
       "Mozilla Extensions",
       "Pointer Events",
+      "Pseudo",
       "Pseudo-classes",
       "Pseudo-elements",
       "Selectors",
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -105,7 +105,7 @@
       "Selectors"
     ],
     "status": "standard",
-    "mdn_url": "https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes"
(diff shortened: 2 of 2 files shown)


ramiy:
I should also note that currently the [CSS Selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) sidebar has no direct link to [Pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes)
and [Pseudo-elements](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements).

This PR adds the "**Pseudo**" group with the links I mentioned above.

ramiy:
CSS selectors - Pseudo

ramiy:
@chrisdavidmills  can you review this PR please?

chrisdavidmills (approved):
LGTM, thanks @ramiy , and sorry that we missed this one for so long.

PR #398, opened 2019-10-29
infra(l10n): Add l10n schema

ExE-Boss:
This&nbsp;adds a&nbsp;JSON&nbsp;schema for&nbsp;files in&nbsp;the&nbsp;`l10n`&nbsp;directory.


teoli2003:
As we want to eliminate this repository, we will never do this. But thanks for the help!

PR #404, opened 2019-12-23
Width height

rachelandrew:
As part of working through experimental features in https://github.com/mdn/sprints/issues/2402 discovered the pages relating to Box Sizing have values listed which are removed from the curent spec (https://github.com/mdn/sprints/issues/2507). This PR updates the formal syntax to match CSS Box Sizing Level 3 for `width`, `min-width`, `height`, `min-height`, `max-width`, `max-height`.

Current spec: https://drafts.csswg.org/css-sizing-3/#preferred-size-properties


chrisdavidmills:
Thanks!

PR #406, opened 2020-01-09
Adds overscroll-behavior block and inline

rachelandrew:
I'm working on https://github.com/mdn/sprints/issues/2635 and so need to add the properties `overscroll-behavior-block` and `overflow-behavior-inline`.

Also updated `nonstandard` to `standard` for the overflow-behavior properties as these have a spec: https://www.w3.org/TR/css-overscroll-1


chrisdavidmills:
LGTM; thanks Rachel!

PR #409, opened 2020-02-27
Added <percentage> value to text-underline-offset and text-decoration…

jmswisher:
…-thickness

For https://bugzilla.mozilla.org/show_bug.cgi?id=1607534
 
See also https://drafts.csswg.org/css-text-decor-4/#underline-offset and https://drafts.csswg.org/css-text-decor-4/#text-decoration-width-property

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -8302,11 +8302,11 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-decoration-style"
   },
   "text-decoration-thickness": {
-    "syntax": "auto | from-font | <length>",
+    "syntax": "auto | from-font | <length> | <percentage> ",
     "media": "visual",
     "inherited": false,
     "animationType": "byComputedValueType",
-    "percentages": "no",
+    "percentages": "1 em",
     "groups": [
       "CSS Text Decoration"
     ],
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
@jmswisher ok, so the Travis build is failing on this one because your percentage value needs to be one of the values defined in https://github.com/mdn/data/blob/88eb4d29d19fba280641df8d0adc40fce300332d/css/properties.schema.json#L43

these are basically the allowed percentage values. The final string shown as a result of those values is taken from the l10n file — https://github.com/mdn/data/blob/master/l10n/css.json — depending on which locale is being viewed.

I see that there's a potential value of `referToElementFontSize` — you could just use this one for now. The spec says that when a percentage size is set for these properties, that percentage is a percentage of 1em, which is basically relative to the element's font-size.

jmswisher:
> these are basically the allowed percentage values. The final string shown as a result of those values is taken from the l10n file — https://github.com/mdn/data/blob/master/l10n/css.json — depending on which locale is being viewed.
> 
> I see that there's a potential value of `referToElementFontSize` — you could just use this one for now. The spec says that when a percentage size is set for these properties, that percentage is a percentage of 1em, which is basically relative to the element's font-size.

There is also `referToSizeOfFont`. I'm not sure of the practical difference between that and `referToElementFontSize`.  

The proper rendering of underlines is shown [here in the spec](https://drafts.csswg.org/css-text-decor-4/#text-line-constancy). The element that an underline is applied to may contain elements with more than one font size, as shown in that example.

chrisdavidmills:
I'd maybe ask the engineer on the bug, to confirm the definition we should use. This is a bit of a fiddly definition.

jmswisher:
Based on chatting with jkew, I'll switch it to `referToElementFontSize`, but it needs more explanation in the textual description of `<percentage>`.

PR #410, opened 2020-02-27
Moved 'from-font' value from text-underline-offset to text-underline-…

jmswisher:
…position, per a spec change

See https://bugzilla.mozilla.org/show_bug.cgi?id=1607308#c1


chrisdavidmills:
Yup, this one looks fine; thanks!

PR #411, opened 2020-03-12
Add 'all' keyword to text-underline-decoration-skip-ink, per bug 1611965

jmswisher:
CSS WG has agreed to ‘all’ in place of ‘always’, per
https://bugzilla.mozilla.org/show_bug.cgi?id=1611965#c4


chrisdavidmills:
Looks OK to me. It makes me slightly uncomfortable that browsers are implementing it and it isn't merged into the spec yet, but I guess it is easy enough to back out if the consensus changes.

PR #413, opened 2020-03-20
feat(css): Update `@import` syntax to CSS Cascade Module Level 4

ExE-Boss:
See <https://drafts.csswg.org/css-cascade-4/#at-import> and [bug 1427715](https://bugzilla.mozilla.org/show_bug.cgi?id=1427715).


chrisdavidmills:
This looks correct, but I'd rather not update this until it's available in browsers. Give me another ping when we are closer to implementation.

Thanks @ExE-Boss .

chole22 (approved):

Rumyra (approved):

PR #414, opened 2020-03-20
Changed initial value of image-orientation to `from-image`

jmswisher:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1607667 and
https://github.com/w3c/csswg-drafts/issues/3799

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5266,7 +5266,7 @@
     "groups": [
       "CSS Images"
     ],
-    "initial": "0deg",
+    "initial": "from-image",
     "appliesto": "allElements",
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
@jmswisher This is no longer happening in 75; see https://bugzilla.mozilla.org/show_bug.cgi?id=1607667#c14

So we'll keep this PR for now, and merge it later once the change actually hits release. I'm afraid this'll also probably mean rolling back some of the other related changes.

Lafi0 (approved):

chrisdavidmills:
Looks like this is now fixed; merging.

PR #417, opened 2020-05-01
Remove from-font value from text-underline-offset

seyedi:
There's not such a value for this property based on [CSSWG](https://drafts.csswg.org/css-text-decor-4/#underline-offset)


seyedi:
@chrisdavidmills Thanks, Yes I removed it from ref page a few minutes ago and for the formal syntax I had to come here :)

chrisdavidmills:
> Yes I removed it from ref page a few minutes ago

Ah ha, good work ;-)

PR #425, opened 2020-09-05
feat(css): add missing '-ms-grid-' properties

kripod:
`-ms-grid-columns` and `-ms-grid-rows` [are aliases](https://css-tricks.com/css-grid-in-ie-css-grid-and-the-new-autoprefixer/) for `grid-template-columns` and `grid-template-rows`, respectively.

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -238,6 +238,38 @@
     "status": "nonstandard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/-ms-flow-into"
   },
+  "-ms-grid-columns": {
+    "syntax": "none | <track-list> | <auto-track-list>",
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
Change looks good. Thanks @kripod !

frenic:
The URLs receives 404. Do you rather change the URLs or do you add redirects?

https://developer.mozilla.org/docs/Web/CSS/-ms-grid-columns
https://developer.mozilla.org/docs/Web/CSS/-ms-grid-rows

chrisdavidmills:
@frenic I've added redirects, and put a note on the property pages to mention the aliases near the top.

frenic:
Nice 👍

PR #426, opened 2020-09-09
Add nonstandard -apple-pay-button property value

squidfunk:
As documented here:
https://developer.mozilla.org/en-US/docs/Web/CSS/appearance


chrisdavidmills:
thanks.

Issue #429, opened 2020-10-05
`-moz-force-broken-image-icon` value should use bracket range notation


pyoor:
According to https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-force-broken-image-icon, the `-moz-force-broken-image-icon` only supports a value of 0 or 1.  As such, the value definition should be changed to:

`<number [0,1]>`

Current definition:
https://github.com/mdn/data/blob/537dde75199d5d8c1904c5e9e12680e7af45afd2/css/properties.json#L921

chrisdavidmills:
Thanks for reporting! Would you be able to submit a PR to fix this?

PR #430, opened 2020-10-06
Use bracket notation for '-moz-force-broken-image-icon'

pyoor:
Fixes https://github.com/mdn/data/issues/429

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -918,7 +918,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/-moz-float-edge"
   },
   "-moz-force-broken-image-icon": {
-    "syntax": "<integer>",
+    "syntax": "<number [0,1]>",
     "media": "visual",
     "inherited": false,
     "animationType": "discrete",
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
Thanks @pyoor . I suppose my question is — can you show me evidence for the square bracket notation part being valid? I mean, `<number>` makes perfect sense, and we have an MDN page for it: https://developer.mozilla.org/en-US/docs/Web/CSS/number

But I've never before come across CSS syntax with the square bracket notation part inside it, and I'm finding it hard to find any information about it. Not to say I don't believe you, but I'd like to learn mroe about it so I can review this PR more effectively ;-)

pyoor:
Bracketed notation was introduced in the [CSS-Values-4 Editors Draft](https://drafts.csswg.org/css-values-4/#numeric-ranges).

pyoor:
I also just realized, this should be `<integer [0,1]>` rather than `<number>` since floating point values are not allowed.

chrisdavidmills:
@pyoor thanks for helping me understand this, and for the further update. This makes sense to me now.

PR #431, opened 2020-10-18
Replace auto -> none for <'max-width'> and <'max-height'>

lahmatiy:
<'max-width'> has `none` value since the beginning, not `auto`.
References:
https://www.w3.org/TR/CSS2/visudet.html#min-max-widths
https://drafts.csswg.org/css-sizing-3/#max-size-properties


chrisdavidmills (approved):
LGTM; thanks @lahmatiy !

PR #434, opened 2020-10-22
Mark size parameter of ray function as mandatory

### LINKED ISSUE #433: Invalid syntax for CSS ray() function
The CSS ray() is defined as follows (via https://drafts.fxtf.org/motion-1/#valdef-offset-path-ray):
`ray( [ <angle> && <size> && contain? ] )`

The current definition marks `<size>` as optional:
https://github.com/mdn/data/blob/e49d197ab3e7fd8b8036e6e7ad46a5aeb1f1a09b/css/properties.json#L6619

This line should be changed to:
`"syntax": "none | ray( [ <angle> && <size> && contain? ] ) | <path()> | <url> | [ <basic-shape> || <geometry-box> ]", `

pyoor:
Fixes #433

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6616,7 +6616,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/offset-distance"
   },
   "offset-path": {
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
LGTM; thanks @pyoor !

PR #438, opened 2020-11-09
Add CSS @property at-role

ramiy:
To fix the scripting error on https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/@property

Warns about the missing formal syntax generated by **{{csssyntax}}**.


chrisdavidmills (approved):
Thanks @ramiy ; your addition looks correct.

PR #439, opened 2020-11-09
Add descriptors to @property at-rule

ramiy:
Source: https://drafts.css-houdini.org/css-properties-values-api-1/#at-property-rule


chrisdavidmills:
I was just looking into fixing the CI breakage on the data repo, and submitted https://github.com/mdn/data/pull/440, which includes these additions. I only noticed your PR afterwards, sorry about that.

PR #446, opened 2020-11-19
Use color type in syntax for border related properties

### LINKED ISSUE #445: Don't reference color property in border-* syntax
Currently, the border related properties include `<'color'>` in their syntax.  This is a reference to the color property which in turn references the `<color>` type.  This should be changed to reference the color type directly as per the spec.

https://drafts.csswg.org/css-logical/#border-shorthands

pyoor:
Fixes #445


chrisdavidmills:
LGTM; thanks @pyoor

PR #448, opened 2020-11-20
Remove numerical bounds from <number> types

Rosuav:
The rendered version of the bounded types displays incorrectly, causing a hard-to-read syntax description. A better fix may be to enhance the transformation tool but this will at least make everything readable.

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -180,7 +180,7 @@
     "syntax": "cross-fade( <cf-mixing-image> , <cf-final-image>? )"
   },
   "cubic-bezier-timing-function": {
-    "syntax": "ease | ease-in | ease-out | ease-in-out | cubic-bezier(<number [0,1]>, <number>, <number [0,1]>, <number>)"
+    "syntax": "ease | ease-in | ease-out | ease-in-out | cubic-bezier(<number>, <number>, <number>, <number>)"
   },
   "deprecated-system-color": {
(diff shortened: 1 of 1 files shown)


Rosuav:
Hmm. The original content seems to be taken straight from the CSSWG https://drafts.csswg.org/css-fonts-4/#font-weight-absolute-values so this simple PR may be a bad idea. I'm not up to the task of fixing the transformation though.

Rosuav:
BTW the faulty render can be seen here: https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#Formal_syntax

papandreou:
As a programmatic user of these specs, I vote for fixing the transformation instead.

lahmatiy:
This is a [CSS bracketed range notation](https://drafts.csswg.org/css-values-4/#numeric-ranges), which was introduced 1-2 years ago. It used in many specs (see [initial PR](https://github.com/w3c/csswg-drafts/pull/3894) affected 7 specs), but not all changes in specs are reflected in `mdn-data` yet. `mdn-data` customers (libraries and tools) are using this notations to validate values.

Therefore, notations should not be removed, since it comes from the specs and more notations will be added to specs / mdn-data in the future.

Rumyra:
Yes we'll be keeping bracketed notation - thanks for the pr tho @Rosuav 👍

Rosuav:
:+1:

Issue #424, opened 2020-08-10
CSS user-select property is incorrectly listed as nonstandard


jmswisher:
This appears to be false, as it is [defined in a spec](https://drafts.csswg.org/css-ui-4/#propdef-user-select), and listed in BCD as on a standards track.

wbamberg:
Can we retire this field from mdn/data? We already maintain this in multiple places (BCD, random macros in the page, "Specifications" sections).

jmswisher:
This issue was prompted because `user-select` is not showing up in the listing constructed by the CSS_Ref macro; so that would need to be updated or killed.

yarusome:
@wbamberg I think this issue is fixed by #449 now. Could this issue be closed?

PR #449, opened 2020-11-23
Change the user-select property's status to standard

abdnh:
This was reported in https://github.com/mdn/kumascript/issues/1246 and https://github.com/mdn/data/issues/424


chrisdavidmills (approved):
Yes, this seems perfectly reasonable. Thanks!

PR #450, opened 2020-11-24
Add `content-visibility` property

frenic:
https://www.w3.org/TR/css-contain-2/#content-visibility

Related https://github.com/frenic/csstype/pull/120

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -4167,6 +4167,22 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/content"
   },
+  "content-visibility": {
+    "syntax": "visible | auto | hidden",
+    "media": "all",
+    "inherited": false,
+    "animationType": "discrete",
+    "percentages": "no",
+    "groups": [
+      "CSS Containment"
+    ],
+    "initial": "visible",
+    "appliesto": "elementsForWhichLayoutContainmentCanApply",
+    "computed": "asSpecified",
+    "order": "perGrammar",
+    "status": "nonstandard",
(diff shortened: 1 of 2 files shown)


artem-malko (approved):

chrisdavidmills (changes requested):
Looks pretty good, thanks @frenic !

One query though — shouldn't the `status` be `standard` ?

frenic:
I choose `nonstandard` because it was a working draft. But really I'm not sure what defines the status in this case. I would love to hear more about how you reason about it.

However, I changed it to `standard`. Thanks!

frenic:
Is this OK to merge now?

chrisdavidmills:
Yes, absolutely. Sorry, I thought I'd done this one!

chrisdavidmills:
And to answer your previous question, generally we say something is a standard if it is somewhere on a standards track, and has some kind of momentum/interest. This page contains our "official" definitions of when to document something, what experimental means, etc: https://developer.mozilla.org/en-US/docs/MDN/Guidelines/Conventions_definitions

It's a tricky question though, and I'm still not sure they are 100% perfect.

frenic:
> And to answer your previous question, generally we say something is a standard if it is somewhere on a standards track, and has some kind of momentum/interest. This page contains our "official" definitions of when to document something, what experimental means, etc: https://developer.mozilla.org/en-US/docs/MDN/Guidelines/Conventions_definitions
> 
> It's a tricky question though, and I'm still not sure they are 100% perfect.

That info help out a lot! Thanks 🙂

PR #451, opened 2020-12-02
Add page-size syntax

frenic:
I noticed `<page-size>` this was missing from the [`size` property](https://github.com/mdn/data/blob/61fa5b22ae6e0474cd96e4ce69fa4b3e15f9ccda/css/at-rules.json#L316)

https://drafts.csswg.org/css-page/#typedef-page-size-page-size


chrisdavidmills (approved):
LGTM; thanks!

PR #456, opened 2020-12-28
Added and updated Japanese translations

mfuji09:
Added Japanese translations for keywords which starts "a".


chrisdavidmills (approved):
LGTM; thanks!

PR #459, opened 2020-12-29
Added and updated Japanese translations in css.json

mfuji09:
Added and updated Japanese translation in css.json
Translated items are id which begins "o" to "r".


mfuji09:
This PR has been replaced with https://github.com/mdn/data/pull/460

PR #464, opened 2021-01-20
adds overflow-clip-margin

rachelandrew:
Adding the data for the `overflow-clip-margin` property, which is part of the Overflow spec https://drafts.csswg.org/css-overflow-3/#overflow-clip-margin

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6967,7 +6967,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Mozilla/CSS/overflow-clip-box"
   },
   "overflow-clip-margin": {
-    "syntax": "<length [0,∞]>",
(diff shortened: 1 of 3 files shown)


pyoor:
Isn't this also missing `<visual-box>` from the syntax?

rachelandrew:
Looks like that has been added to the spec since I made this PR @pyoor 

https://github.com/w3c/csswg-drafts/commit/337369a30d8e3a375281a0a876f5b6eb4987642f

thanks for spotting it :)

pyoor:
@rachelandrew You may also want to add `<visual-box>` to types.json just as a reference.  I'm not sure what the procedure is for linking between types in mdn syntax data.

chrisdavidmills (approved):

PR #465, opened 2021-01-26
Wrap :dir arguments in group brace

pyoor:
The syntax for the `:dir` pseudo class is incorrect.  The syntax is intended to reflect that `:dir` can accept either the `rtl` or `ltr` arguments.  However, the current syntax actually denotes two definitions: 
`:dir( ltr` and `rtl )`

The correct definition is:
`:dir( [ ltr | rtl ] )`

### DIFF
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -180,7 +180,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:defined"
   },
   ":dir": {
-    "syntax": ":dir( ltr | rtl )",
+    "syntax": ":dir( [ ltr | rtl ] )",
     "groups": [
       "Pseudo-classes",
       "Selectors"


chrisdavidmills (commented):
Are you sure this is needed? I thought just | was enough to indicate "either, or".

Can you find a listing of the syntax in the spec for `:dir`? I had a quick look, but couldn't find anything.

pyoor:
It is because the `|` in `:dir( ltr | rtl )` does denote "either, or" but it's not limited to just the function values.  Instead, `:dir( ltr | rtl )` represents either `:dir( ltr` or `rtl )`.

The spec that defines `:dir()` displays the syntax as `:dir(ltr)` and `:dir(rtl)` which could be used here as well in the form of `:dir(ltr) | :dir(rtl)` instead of wrapping the arguments in a group (i.e. `:dir( [ ltr | rtl ] )`

chrisdavidmills:
Ah, of course — thank you for clarifying that.

In that case, yes, let's get it merged. Thanks @pyoor !

PR #466, opened 2021-01-29
adding color-scheme

rachelandrew:
Adding the `color-scheme` property

https://drafts.csswg.org/css-color-adjust/#color-scheme-prop

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -3955,7 +3955,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color-adjust"
   },
   "color-scheme": {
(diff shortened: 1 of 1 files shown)


css/properties.json
```diff
@@ -3954,6 +3954,22 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color-adjust"
   },
+  "color-scheme": {
+    "syntax": "normal | [ light | dark | <custom-ident>]+",
```
lahmatiy, line 3958: It would be nice to add a space before `]`

chrisdavidmills (approved):

PR #467, opened 2021-01-29
List style image

rachelandrew:
Working on https://github.com/mdn/content/issues/1728

Updating `list-style-image` as per the spec: https://drafts.csswg.org/css-lists-3/#propdef-list-style-image

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5801,7 +5801,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/list-style"
   },
   "list-style-image": {
-    "syntax": "<> | none",
+    "syntax": "<image> | none",
     "media": "visual",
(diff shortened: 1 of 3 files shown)


css/properties.json
```diff
@@ -5801,7 +5801,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/list-style"
   },
   "list-style-image": {
-    "syntax": "<url> | none",
+    "syntax": "<> | none",
```
lahmatiy, line 5804: Should be `<image>` instead of `<>`

rachelandrew: that's odd, not quite sure how I managed to remove that, fixed anyway.

chrisdavidmills (approved):
OK, let's get this one in.

PR #470, opened 2021-02-01
adding ::target-text

rachelandrew:
Adding the ::target-text pseudo. I'll be creating the rest of the docs.

Spec: https://drafts.csswg.org/css-pseudo/#selectordef-target-text
Chrome Status: https://www.chromestatus.com/feature/5689463273422848


chrisdavidmills:
Thanks @rachelandrew !

PR #471, opened 2021-02-04
new value in percentages enum

Konrud:
"logicalHeightOrWidthOfContainingBlock" to represent both values for percentage.
 It's needed for the other commit for the `inset` CSS property.


chrisdavidmills (approved):
Makes sense to me; thanks @Konrud !

PR #472, opened 2021-02-04
new value "logicalHeightOrWidthOfContainingBlock"

Konrud:
Values have been added for both English and Russian.
 It's needed for the other commit for the `inset` CSS property.


chrisdavidmills (approved):
@Konrud I've made a very small edit to the English text, but this basically looks fine too. Thanks!

PR #474, opened 2021-02-05
Replace positive-integer types with ranged notation

pyoor:
The type `<positive-integer>` is not a standard terminal type and should be replaced with ranged notation.

See also https://github.com/w3c/csswg-drafts/issues/355

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -273,7 +273,7 @@
     "syntax": "<length-percentage>"
   },
   "fixed-repeat": {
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
OK, cool.

SebastianZ:
Be aware that the CSSSyntax macro isn't able to parse this syntax correctly at the moment. See https://github.com/mdn/kumascript/issues/1441.

I wanted to fix that, though I couldn't get [Yari](https://github.com/mdn/yari) to run on my machine yet.

Sebastian

PR #475, opened 2021-02-18
Fix fit-content syntax

playhardgopro:
`fit-content` is a function but we can use it without parenthesises and arguments
https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content

So, I suggest these changes to the syntax

Thanks!

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5338,7 +5338,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/hanging-punctuation"
   },
   "height": {
-    "syntax": "auto | <length> | <percentage> | min-content | max-content | fit-content [ (<length-percentage>) ]?",
+    "syntax": "auto | <length> | <percentage> | min-content | max-content | fit-content | fit-content(<length-percentage>)",
     "media": "visual",
     "inherited": false,
     "animationType": "lpc",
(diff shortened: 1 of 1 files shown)


css/properties.json
```diff
@@ -5338,7 +5338,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/hanging-punctuation"
   },
   "height": {
-    "syntax": "auto | <length> | <percentage> | min-content | max-content | fit-content(<length-percentage>)",
+    "syntax": "auto | <length> | <percentage> | min-content | max-content | fit-content [ (<length-percentage>) ]?",
```
lahmatiy, line 5341: This incorrect, since instead `fit-content(` function-token, two tokens will be produced (ident `fit-content` and `(`). I suppose it should be:
```suggestion
    "syntax": "auto | <length> | <percentage> | min-content | max-content | fit-content | fit-content(<length-percentage>)",
```

playhardgopro: ok, I thought it's the same but less code\
so question mark `?` is for optional group here
for example:
`composes: '<custom-ident> [ from [ <string> | global ] ]?',`

chrisdavidmills:
Let me know when you're happy with this, and I'll merge it.

css/properties.json
lahmatiy: Yeah, `?` combinator stands for 0 or 1 occurrences.
However there is a difference between `foo(` and `foo (`. In case it should be a function, any whitespaces should be omitted between keyword and parenthesis. See [function-token](https://www.w3.org/TR/css-syntax-3/#function-token-diagram) definition.
So grammar in this PR allows `fit-content ( 100px )` (not a function), and disallows `fit-content(100px)` (a function).

lahmatiy: Sorry, I missed you'd changed definitions. Now it looks OK.

lahmatiy:
LGTM

chrisdavidmills:
Merged; thanks a lot!

PR #481, opened 2021-03-29
update ruby-position

rachelandrew:
The syntax and initial value of `ruby-position` has changed in the spec: https://www.w3.org/TR/css-ruby-1/#rubypos

This PR updates these as part of https://github.com/mdn/content/issues/3457


chrisdavidmills (approved):
LGTM; thanks @rachelandrew !

PR #489, opened 2021-06-10
Update `:host()`, `:host-context()`, and `::slotted()` syntax to `compound-selector`

aomarks:
The `:host()`, `::host-context()`, and `::slotted()` pseudo-classes now take a `compound-selector`, instead of a `compound-selector-list`.

Change: https://github.com/w3c/csswg-drafts/pull/5673/files
Discussion: https://github.com/w3c/csswg-drafts/issues/2158
Latest spec sections:
https://drafts.csswg.org/css-scoping/#selectordef-host-function
https://drafts.csswg.org/css-scoping/#selectordef-host-context
https://drafts.csswg.org/css-scoping/#selectordef-slotted

### DIFF
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -297,7 +297,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:has"
   },
   ":host()": {
-    "syntax": ":host( <compound-selector-list> )",
+    "syntax": ":host( <compound-selector> )",
     "groups": [
       "Pseudo-classes",
       "Selectors"
@@ -306,7 +306,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:host()"
   },
   ":host-context()": {
-    "syntax": ":host-context( <compound-selector-list> )",
+    "syntax": ":host-context( <compound-selector> )",
     "groups": [
       "Pseudo-classes",
       "Selectors"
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Hi @aomarks and thanks for this PR. I'm sorry you had to wait so long for a review (I was only just now alerted to it). This looks good. Thank you for the update. 👍

aomarks:
> Hi @aomarks and thanks for this PR. I'm sorry you had to wait so long for a review (I was only just now alerted to it). This looks good. Thank you for the update. 👍

Thanks @ddbeck !

For future reference, would it be a good practice to tag you, or somebody else you might recommend, on PRs like this (maybe after a few days have passed?). Don't want to bug anyone unnecessarily, just looking to find the best process.

I'm also curious how frequently the changes in this repo make their way out to the live MDN site?

Thanks!

ddbeck:
@aomarks Tagging me is fine for now, yes. We're working on getting a better handle on our PR review process. I hope that, in the coming weeks/months, we'll be developing some more formal processing and review processes. Until recently, we were severely understaffed and hadn't been able to make headway on that project, but it's increasingly our focus now.

> I'm also curious how frequently the changes in this repo make their way out to the live MDN site?

Your changes should be live now (though let me know if you see anything out of place). This repo is occasionally packaged into a new release (which we do more or less as needed) and then deployed a day or two later to MDN. MDN itself is aggressively cached in the CDN, so it sometimes takes a day or two for the current deployment to reach readers.

PR #494, opened 2021-07-25
[WIP] fix(css): update `mask-repeat` and `mask-position` according to spec

### LINKED ISSUE #351: `mask-repeat` has incorrect initial value
Spec: https://drafts.fxtf.org/css-masking-1/#the-mask-repeat

Right value: `no-repeat`

sigveio:
This pull request aims to update the initial value of `mask-repeat` and `mask-position` to be in line with the current spec.

Specs: 
https://drafts.fxtf.org/css-masking/#the-mask-repeat
https://drafts.fxtf.org/css-masking/#the-mask-position

Additional context:
w3c/csswg-drafts/issues/599
w3c/fxtf-drafts/issues/111


sigveio:
Going over the spec for this again, I noticed that some of the other JSON properties for these two could also use updating. So I'm marking it as a WIP for now.

This is the definition of `appliesto` in the latest spec for these elements:

> All elements. In SVG, it applies to container elements excluding the defs element, all graphics elements and the use element

Which deviates from the `allElementsSVGContainerElements` used in the JSON currently.

What would be a suitable replacement? 

Attempting to break it down:

`All elements.` -> allElements
`In SVG, it applies to container elements` -> SVGContainerElements
`excluding the defs element, all graphics elements and the use element` -> ExcludingDefsGraphicsAndUse

Result: 

`allElementsSVGContainerElementsExcludingDefsGraphicsAndUse`

or (perhaps slightly clearer)

`allElementsAndSVGContainerElementsExcludingDefsGraphicsAndUse`

Thoughts? 

`Excluding` (more in line with text in spec) or `Except` (more in line with existing schema entries)?

css/properties.json
```diff
@@ -6381,7 +6381,7 @@
     "groups": [
       "CSS Masking"
     ],
-    "initial": "center",
+    "initial": "0% 0%",
```
skyclouds2001, line 6384: this is already landing in https://github.com/mdn/data/commit/2987ea9827a8e8a29cbae0ed0a6bf02dffab3301

css/properties.json
```diff
@@ -6397,7 +6397,7 @@
     "groups": [
       "CSS Masking"
     ],
-    "initial": "no-repeat",
+    "initial": "repeat",
```
skyclouds2001, line 6400: this is already landing in https://github.com/mdn/data/commit/b36469e253ba551ba2a5a9738b814522309e5f6a

skyclouds2001:
> Going over the spec for this again, I noticed that some of the other JSON properties for these two could also use updating. So I'm marking it as a WIP for now.
> 
> This is the definition of `appliesto` in the latest spec for these elements:
> 
> > All elements. In SVG, it applies to container elements excluding the defs element, all graphics elements and the use element
> 
> Which deviates from the `allElementsSVGContainerElements` used in the JSON currently.
> 
> What would be a suitable replacement?
> 
> Attempting to break it down:
> 
> `All elements.` -> allElements `In SVG, it applies to container elements` -> SVGContainerElements `excluding the defs element, all graphics elements and the use element` -> ExcludingDefsGraphicsAndUse
> 
> Result:
> 
> `allElementsSVGContainerElementsExcludingDefsGraphicsAndUse`
> 
> or (perhaps slightly clearer)
> 
> `allElementsAndSVGContainerElementsExcludingDefsGraphicsAndUse`
> 
> Thoughts?
> 
> `Excluding` (more in line with text in spec) or `Except` (more in line with existing schema entries)?

Personnally, both is ok, as this does not affect user

bsmth:
After updating from main, there's no changes, so the linked commits have resolved this. I'm going to close this, thank you!

PR #496, opened 2021-07-28
fix(css): update `color` and `opacity` to css-color-4 spec

### LINKED ISSUE #495: Initial value of `color`
Combing through drafts from the CSS Working Group I noticed that as of [~january last year](https://github.com/w3c/csswg-drafts/issues/4608) it was decided that the `initial` value of `color` would use `CanvasText`. This was merged in w3c/csswg-drafts/pull/4714 and has been published in the last two versions of the working draft for [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4/#the-color-property) (css-color-4).

I tried to trace the status with the major browser platforms, and I can see mozilla/gecko and chronium appears to have fixed [[mozilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1590894), [chronium](https://bugs.chromium.org/p/chromium/issues/detail?id=1015968)] and shipped [[mozilla](https://hg.mozilla.org/mozilla-central/rev/3f2a791d050a), [chronium](https://chromium-review.googlesource.com/c/chromium/src/+/1864378)]. There doesn't appear to be any public signals from WebKit, but judging from the [web-platform-tests](https://wpt.fyi/results/css/css-color/color-initial-canvastext.html?label=master&label=stable&aligned) it looks good there too.

I'm thinking that it could be time to update the `initial` value of `color` in the MDN CSS property data to `CanvasText` instead of the special keyword "variesFromBrowserToBrowser". As it's no longer flagged as "depends on user agent" in css-color-4.

Besides keeping the [MDN developer documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/color#formal_definition) for `color` (which already [references css-color-4](https://developer.mozilla.org/en-US/docs/Web/CSS/color#specifications)) up-to-date, it would also be useful for build tools that rely on the data to optimise or minimise CSS.

But I thought I would ask for feedback to see if there are arguments against this change before I submit a PR. =)

Also; for future reference - since I wouldn't mind to continue helping you keep the data up-to-date - are there any official criteria or common practices for MDN data in terms of determining when it's deemed safe to update a property? Am I correct in thinking that you generally follow the current working draft and accepted browser support/implementation? Perhaps you could shed some light on that, @rachelandrew?

sigveio:
## Summary

`color` and `opacity` do not currently match the spec, CSS Color Module Level 4 (css-color-4).

This pull request aims to update them.

Closes #495

### `color`

CSSWG have decided to use use `canvastext` as the `initial` value of `color` - making it consistent across user agents.

I'm also updating the various other JSON properties of `color` to be in line with the spec.

#### Additional resources/context:

Formal definition: https://www.w3.org/TR/css-color-4/#the-color-property
Resolution: w3c/csswg-drafts/issues/4608
Merged draft PR: w3c/csswg-drafts/pull/4714
WPT: [color-initial-canvastext.html](https://wpt.fyi/results/css/css-color/color-initial-canvastext.html?label=master&label=stable&aligned)

----

### `opacity`

This was also out of sync, so as with `color` i updated the JSON to be in line with the spec.

Here it's worth noting that `percentages` changed to `map to the range [0,1]` and I could not find a corresponding entry in the schema. Nor any documentation on how the schema is constructed.

I therefore added a suggestion myself as `mapToRange0To1`. You may want to double-check internally that adding to the schema like this is the correct approach, as I am not aware if it should map/correspond to something in particular.

#### Additional resources/context:

Formal definition: https://www.w3.org/TR/css-color-4/#transparency

----

## Notes

I checked the git blame and could not see any regression against previous issues with these changes.


css/properties.json
```diff
@@ -3940,15 +3940,15 @@
...
+    "animationType": "byComputedValueType",
     "percentages": "no",
     "groups": [
       "CSS Color"
     ],
-    "initial": "variesFromBrowserToBrowser",
-    "appliesto": "allElements",
-    "computed": "translucentValuesRGBAOtherwiseRGB",
-    "order": "uniqueOrder",
+    "initial": "canvastext",
```
sigveio, line 3948: I opted for the lowercased "canvastext" to be in line with for example "currentcolor" in the file

css/properties.schema.json
```diff
@@ -49,6 +49,7 @@
         "logicalHeightOfContainingBlock",
         "logicalWidthOfContainingBlock",
         "logicalHeightOrWidthOfContainingBlock",
+        "mapToRange0To1",
```
sigveio, line 52: When adding this new percentages enum representing `map to the range [0,1]`, I dropped 'the' when concatenating it and wrote the range in the same fashion as existing entries. See for example `specifiedValueNumberClipped0To1`

As noted in the PR body, it might be worth triple checking if this addition to the schema is correct.

l10n/css.json
```diff
@@ -970,6 +970,9 @@
     "ja": "両方の値が length の場合は、length 値として補完されます。両方の値がパーセント値の場合は、パーセント値として補完されます。それ以外の場合は、両方の値が calc() 関数にコンバートされ、length とパーセント値の合計になります (または各値が 0)。そして、これらの <code>calc()</code> 関数は、それぞれ半分ずつ補完された実数を持ちます",
     "ru": " когда оба значения являются длинами, они интерполируются как длины; когда оба значения являются процентами, они интерполируются как проценты; в остальных случаях, оба значения преобразуются функцией <code>calc()</code> в сумму длины и процента (каждый из них может быть равен нулю), а затем эта функция интерполирует каждую половину как вещественные числа."
   },
+  "mapToRange0To1": {
```
sigveio, line 973: Since I added a new entry to the schema, I decided to add a corresponding `l10n` entry for `en-US`. It's possible that this is superfluous, since it does not add localization beyond the base language for MDN. 

My rationale for adding it anyway:

- It helps document `mapToRange0To1` with a human readable version
- It makes it more accessible for others to see where translations are missing and add their native language

sigveio:
@rachelandrew if you have time to give me feedback on this PR, at your convenience, that would be appreciated. Or alternatively, could you be so kind to request a review from someone else for me please? As I don't know who best to ask in that case.

I believe there is much more that needs updating, and I'd be happy to help. But unfortunately I stall without feedback on whether my thinking around the schema and such is correct.

Rumyra:
@sigveio this all looks good to me. I think the addition of `mapToRange0to1` seems appropriate 👍

PR #503, opened 2021-09-02
Fix -moz-force-broken-image-icon formal syntax since rendering appears broken

SphinxKnight:
Currently, [the page for -moz-force-broken-image-icon](https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-force-broken-image-icon#formal_syntax) displays this for the formal syntax: 
image
where we can see an unmatched angle bracket.

This change should be sufficient to fix this.

Note; reported on Twitter https://twitter.com/nhoizey/status/1433332949236264961

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -950,7 +950,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/-moz-float-edge"
   },
   "-moz-force-broken-image-icon": {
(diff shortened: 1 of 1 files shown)


teoli2003 (approved):

teoli2003:
Thanks! It will be updated the next time mdn/data is generated.

PR #512, opened 2021-10-01
Add entry for input-security property

### LINKED ISSUE #511: Add an entry for `input-security` CSS property.
https://drafts.csswg.org/css-ui-4/#input-security

lukewarlow:
Adds an entry for https://drafts.csswg.org/css-ui-4/#input-security

Not sure if the `appliesto` value is correct. Should I add a sensitiveTextInputs value to the enum and use that instead?

Related to https://github.com/mdn/content/issues/9422

Closes #511


Rumyra:
Thanks so much @lukewarlow :) In a word yes, but as we've been a bit lax reviewing data prs, I just went ahead and added for you 👍

PR #514, opened 2021-10-26
Updating for @layer and layer()

Rumyra:
OK this is my first pass of updating the data repo, so pls be kind.

There is an @layer rule and also a layer() function - which I hope I have added successfully. One part is the @import rule which harnesses the layer() function. I'm not sure how to update for that syntax, feedback welcome! Thnx

https://github.com/mdn/content/issues/9369


css/syntaxes.json
```diff
@@ -368,6 +368,9 @@
   "keyframe-selector": {
     "syntax": "from | to | <percentage>"
   },
+  "layout()": {
+    "syntax": "layout( <layer-name> )"
+  },
```
Rumyra, line 373:
```suggestion
  "layer()": {
    "syntax": "layer( <layer-name> )"
  },
```
```suggestion
  "layout()": {
    "syntax": "layout( <layer-name> )"
  },
```

css/at-rules.json
```diff
@@ -292,6 +292,17 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@keyframes"
   },
+  "@layer": {
+    "syntax": "@layer <layer-name> {\n  <stylesheet>\n}",
+    "interfaces": [
+      "CSSImportRule"
```
wbamberg, line 298: According to the docs (https://github.com/mdn/data/blob/master/css/at-rules.md#structure-for-simple-at-rules) this property is "are the CSSOM interfaces that belong to the at-rule" I'm not sure what "belong to" means here, but I think there's a good case for not including an `interfaces` property here.

(I don't know anywhere that this property is used in our docs though.)

css/syntaxes.json
```diff
@@ -368,6 +368,9 @@
   "keyframe-selector": {
     "syntax": "from | to | <percentage>"
   },
+  "layer()": {
+    "syntax": "layer( <layer-name> )"
+  },
```
wbamberg, line 373: I couldn't find anywhere this is defined in a spec, can you point me to it? It looks as if this participates in the definition of `@import` (https://www.bram.us/2021/09/15/the-future-of-css-cascade-layers-css-at-layer/#cascade-layers--create), so maybe the syntax for `@import` should be updated to refer to it? But I can't find any reference to `layer()` in the spec for `@import`, or in our page.

wbamberg (changes requested):
Sorry to take so long to review this @Rumyra . I'm not very good on this stuff myself but have tried!

In general I'm not sure how or where this is used in our docs, apart from the syntax, (that is used in the formal syntax definitions).

css/at-rules.json
Rumyra: Yes I actually think this is a mistake and should be these two interfaces https://drafts.csswg.org/css-cascade-5/#the-csslayerblockrule-interface which I've changed

css/syntaxes.json
Rumyra: It's the function used to declare a layer within the `@import` rule https://drafts.csswg.org/css-cascade-5/#layer-declaration - I've updated the syntax for the this as well - and yes the docs need updating but I'll get to that in due course 👍

css/at-rules.json
```diff
@@ -292,6 +292,17 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@keyframes"
   },
+  "@layer": {
+    "syntax": "@layer <layer-name> {\n  <stylesheet>\n}",
```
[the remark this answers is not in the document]
Rumyra: I think we can go with your suggestion for the `@layer` syntax. I've added `<layer-name>` to syntaxes

PR #520, opened 2021-12-01
fix -webkit-tap-highlight-color property's 'inherited' data

AnilSeervi:
false -> true, as it is an inherited property according to its [specs](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/AdjustingtheTextSize/AdjustingtheTextSize.html#//apple_ref/doc/uid/TP40006510-SW5)

Closes mdn/content#10886


Rumyra:
Thanks @AnilSeervi 👍

PR #548, opened 2022-01-25
Adding `animation-timeline` property

Rumyra:


ddbeck:
The failure here is that `better-ajv-errors` got upgraded improperly; see [the release notes](https://github.com/atlassian/better-ajv-errors/blob/main/CHANGELOG.md#breaking-changes). I think the `require` needs to change on this line:

https://github.com/mdn/data/blob/main/test/lint.js#L4

Then it should either fail with an actual message or pass.

PR #557, opened 2022-02-10
Fix initial value of appearance

### LINKED ISSUE #554: Appearance initial value is "none", not "auto"
Hi, 
it seems data for appearance is wrong, both firefox and chrome have implemented initial as none and w3c specified none

https://www.w3.org/TR/2021/WD-css-ui-4-20210316/#propdef-appearance

teoli2003:
Initial value is `auto` and no more `none`.

Fixes #554

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -1972,7 +1972,7 @@
     "groups": [
       "CSS Basic User Interface"
     ],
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

PR #558, opened 2022-02-10
Fix initial values of background-position-{x,y}

### LINKED ISSUE #555: Property background-position-x and -y not quite correct
w3c and both firefox and chrome show 0% for both background-position-x and background-position-y

While technically left and top is also correct, I think we should stick with the spec (yeah I found it because the logic I am writing requires an exact match with the browser value)

https://drafts.csswg.org/css-backgrounds/#propdef-background-position

teoli2003:
Fixes #555

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -2252,7 +2252,7 @@
     "groups": [
       "CSS Backgrounds and Borders"
     ],
-    "initial": "left",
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

PR #564, opened 2022-02-25
Add missed <layer-name> syntax

lahmatiy:
https://www.w3.org/TR/css-cascade-5/#typedef-layer-name


teoli2003 (approved):
This is correct. Thank you!

PR #565, opened 2022-02-25
Add missed <single-animation-timeline> and <timeline-name>

lahmatiy:
https://drafts.csswg.org/css-animations-2/#typedef-single-animation-timeline
https://drafts.csswg.org/css-animations-2/#typedef-timeline-name


teoli2003 (approved):
Thank you. This will fix the formal syntax in https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timeline

PR #568, opened 2022-03-03
Add lab() & lch() to color

NiedziolkaMichal:
Functions [lab](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/lab()) & [lch](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/lch()) are already defined in [color_value page](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) and supported on Safari. I see that [specification](https://www.w3.org/TR/css-color-4/#color-syntax) also mentiones &lt;oklab()> | &lt;oklch()>, but they aren't defined anywhere in MDN, should I add them too?

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -102,7 +102,7 @@
     "syntax": "<url>"
   },
   "color": {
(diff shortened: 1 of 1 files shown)


teoli2003 (approved):

teoli2003:
I would wait for some implementation before adding oklab & co.

teoli2003:
Thanks a lot!

Looks good to me.

PR #571, opened 2022-03-08
Adds repeating-conic-gradient()

NiedziolkaMichal:
I added missing repeating-conic-gradient() defined in [CSS Images 4](https://www.w3.org/TR/css-images-4/#repeating-gradients).
Syntax is copy paste from conic-gradient().


teoli2003 (approved):
👍

teoli2003:
Note that mdn/data is published about once a month, so the changes won't be visible on MDN until then.

NiedziolkaMichal:
That's good to know, thank you

Issue #567, opened 2022-03-01
Issue with "outline-style": Formal Syntax incorrectly implies that all border-style values can be used.


NiedziolkaMichal:
MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/outline-style

#### What information was incorrect, unhelpful, or incomplete?
Formal syntax
auto | &lt;'border-style'&gt;

#### What did you expect to see?
Formal syntax
auto | &lt;outline-line-style&gt;

According to [specification](https://w3.org/TR/css-ui-4/#typedef-outline-line-style) outline-style property doesn't accept value "hidden", while border-style does. I believe it will be better to change Formal Syntax and mention how outline-line-style is different from border-style.

<details>
<summary>MDN Content page report details</summary>

* Folder: `en-us/web/css/outline-style`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/outline-style
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/outline-style/index.md
* Last commit: https://github.com/mdn/content/commit/26f50bdaeb1228f458d21dbdf7dc110387b20daa
* Document last modified: 2021-08-12T23:59:09.000Z

</details>

yarusome:
@teoli2003 Since the transition of `{{CSSSyntax}}` to webref, I guess this issue can now be closed (as well as the PR).

teoli2003:
There are third-party users of mdn/data.

PR #574, opened 2022-03-15
fix(css): update "outline-style" syntax

timeowilliams:
Formal Syntax incorrectly implies that all border-style values can be used. Fix https://github.com/mdn/data/issues/567

See for more info: https://www.w3.org/TR/css-ui-4/#typedef-outline-line-style

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6948,7 +6948,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/outline-offset"
   },
   "outline-style": {
-    "syntax": "auto | <'border-style'>",
+    "syntax": "auto | <'outline-line-style'>",
     "media": [
       "visual",
       "interactive"


css/properties.json
```diff
@@ -6948,7 +6948,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/outline-offset"
   },
   "outline-style": {
-    "syntax": "auto | <'border-style'>",
+    "syntax": "auto | <'outline-line-style'>",
```
teoli2003, line 6951: We also need to have outline-line-style defined somewhere.

teoli2003 (changes requested):

css/properties.json
timeowilliams: Hey @teoli2003 - for getting this resolved, I'm not finding any MDN URL specified for this. I'm going to head over to the mdn/content repo and submit a PR for this. 

Is this what we had in mind?

timeowilliams: Or was it something along the lines of a similiar example, [outline-width](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-width#line-width)

teoli2003: I think we should list the valid values and not use  <'outline-line-style'>". It seems not useful to create a whole type page for something used in one place.

caugner: @timeowilliams Do you still want to add the values?

Basically ([source](https://www.w3.org/TR/css-ui-4/#typedef-outline-line-style)):

> `<outline-line-style> accepts the same values as <line-style> (CSS Backgrounds 3 §3.2 Line Patterns: the border-style properties) with the same meaning, except that hidden is not a legal outline style. In addition, the outline-style property accepts the value auto.`

And `line-style` accepts these values ([source](https://www.w3.org/TR/css-backgrounds-3/#typedef-line-style)):

> `<line-style> = none | hidden | dotted | dashed | solid | double | groove | ridge | inset | outset`

bsmth:
I think this is fixed in the merged PR above, so I'm going to close this. Thanks a lot!

PR #581, opened 2022-04-21
plus-lighter in mix-blend-mode

NiedziolkaMichal:
plus-lighter value is present in [specification](https://drafts.fxtf.org/compositing/#mix-blend-mode) and [implemented](https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode#browser_compatibility) on chromium and firefox.

Specification also includes plus-darker value, but it is not supported anywhere, so I think it shouldnt be added yet.


teoli2003 (approved):
I agree

PR #583, opened 2022-05-26
chore: update publish-release workflow

schalkneethling:
Add required `target-repo` input.
Update required as a result of mdn/workflows#55


teoli2003:
I missed this one. Should we merge it?

schalkneethling:
> I missed this one. Should we merge it?

Yup, that would solve the problem you ran into recently.

PR #589, opened 2022-07-11
fix(css): correct two syntax examples in selectors

tomrav:
`:playing` and `::-ms-expand` have syntax examples from other selector examples.


teoli2003 (approved):
Thank you, good catch! Likely an cut&paste error.

teoli2003:
Thanks a lot!

Looks good to me.

PR #591, opened 2022-08-06
Add missed syntaxes: <lab()>, <lch()>, <ratio> and <reversed-counter-name>

lahmatiy:
Missed syntaxes are used in properties and other syntax.

- `<lab()>`: https://www.w3.org/TR/css-color-4/#specifying-lab-lch
- `<lch()>`: https://www.w3.org/TR/css-color-4/#specifying-lab-lch
- `<ratio>`: https://www.w3.org/TR/css-values-4/#ratios
- `<reversed-counter-name>`: https://drafts.csswg.org/css-lists/#typedef-reversed-counter-name

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -377,11 +377,17 @@
   "keyframe-selector": {
     "syntax": "from | to | <percentage>"
   },
+  "lab()": {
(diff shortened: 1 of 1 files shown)


teoli2003 (approved):
👍

PR #592, opened 2022-08-08
Add syntaxes to CSS trigonometric functions

ramiy:
Part of the documentation work on [CSS trigonometric functions](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions#trigonometric_functions).

[Safari/webkit already supports](https://webkit.org/blog/12445/new-webkit-features-in-safari-15-4/) those CSS functions. [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1190444) will follow soon.

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -3,7 +3,7 @@
     "syntax": "xx-small | x-small | small | medium | large | x-large | xx-large | xxx-large"
   },
   "acos()": {
-    "syntax": "acos( [ <number> | <angle> ] )"
+    "syntax": "acos( <calc-sum> )"
   },
   "alpha-value": {
     "syntax": "<number> | <percentage>"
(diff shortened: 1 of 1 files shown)


css/syntaxes.json
```diff
@@ -677,6 +692,9 @@
   "side-or-corner": {
     "syntax": "[ left | right ] || [ top | bottom ]"
   },
+  "sin()": {
+    "syntax": "sin( [ <number> | <angle> ] )"
```
teoli2003, line 696: Why `<number>` and not something like `<calc-sum>`?

This syntax makes `sin(2 * 0.5)` invalid.

ramiy: Because this is what the spec says https://www.w3.org/TR/css-values-4/#trig-funcs

ramiy: My mistake, found this https://www.w3.org/TR/css-values-4/#calc-syntax

You are correct. I'll fix the PR.

ramiy:
@chrisdavidmills @rachelandrew @lahmatiy @ExE-Boss please review and merge.

teoli2003 (approved):
👍

PR #593, opened 2022-08-09
Add syntaxes to CSS math functions

ramiy:
This PR adds missing CSS math functions syntaxes.

Source: https://www.w3.org/TR/css-values-4/#calc-syntax

Related #592


teoli2003 (approved):
👍

PR #595, opened 2022-08-22
Align <perspective()> and <scale*()> syntaxes with CSS Transforms 2

lahmatiy:
The PR aligns syntaxes with [CSS Transforms 2](https://www.w3.org/TR/css-transforms-2/) which already supported by modern browsers:
- `perspective(none)`
- the scale functions now support `<percentage>`


teoli2003 (approved):
👍

PR #596, opened 2022-08-22
Fix some syntaxes from CSS Grid Layout

lahmatiy:
Align some syntaxes to [CSS Grid Layout 1](https://www.w3.org/TR/css-grid-1/)

Changes:
- Replaced `<length> | <percentage>` -> `<length-percentage>`
- Removed `fit-content()` as never used, it's always defined in place.


teoli2003 (approved):
Thank you!

PR #602, opened 2022-09-06
Add data for math-shift/math-depth

fred-wang:
Manually copy data from webref:
https://github.com/w3c/webref/blob/main/ed/css/mathml-core.json#L39

See also:
https://github.com/mdn/content/issues/18641
https://github.com/mdn/content/pull/20314


teoli2003 (approved):
👍

PR #606, opened 2022-09-26
properties-schema: elementsForWhichLayoutContainmentCanApply

hamishwillee:
This is to allow addition of the contain-intrinsic-width, contain-intrinsic-height, contain-intrinsic-block-size, contain-intrinsic-inline-size in https://w3c.github.io/csswg-drafts/css-sizing-4/#intrinsic-size-override

Issue #585, opened 2022-06-08
Making the CSS property `appliesto` values more specific


captainbrosset:
It feels to me like the `appliesto` property of the CSS properties data isn't defined very specifically at the moment.

For context, here is a concrete example: 
https://developer.mozilla.org/en-US/docs/Web/CSS/align-items#formal_definition

MDN says that `align-items` applies to all elements ([data is here](https://github.com/mdn/data/blob/main/css/properties.json#L1732)). While true in the sense that it won't break if you apply it to any element, it actually won't do anything unless the element is not a grid container or a flex container.

Contrast this with this other example:
https://developer.mozilla.org/en-US/docs/Web/CSS/grid-auto-columns#formal_definition

MDN says that `grid-auto-columns` only applies to grid containers, which is true. But also, it won't cause any problem if you apply it to any other type of containers.

I would love if `appliesto` was more specific. There seems to be a lot of different values that the `appliesto` enum can take: https://github.com/mdn/data/blob/main/css/properties.schema.json#L162
and using the most specific one has advantages: it makes it easier for web authors to know why a property they're using isn't doing anything.

As an example, if you're confused about flexbox and you're trying to align items using `align-content` but you did not specify `flex-wrap:wrap`, then the property won't have any effects.

In this specific case, the data is actually very helpful. MDN says that `align-content` only applies to multi-line flex containers: https://developer.mozilla.org/en-US/docs/Web/CSS/align-content#formal_definition
Although it's not entirely correct because it also applies to grid containers.

So I have two questions in this issue:

1. Does everyone agree that the `appliesto` property should be as specific as possible for all properties?
2. Is there interest in expanding the `appliesto` enum to make it able to cover more cases? For example, there is no `gridAndFlexContainer` value now, but it would be useful in the `align-content` case. In fact, it might become useful, at some point, to allow more complex values than just strings. For instance: `gridContainer AND flexContainer` seems like it would help.

As background to this, in the past I worked on this inactive-css repo with the hope to create a reusable dataset for exactly this case: https://github.com/captainbrosset/inactive-css
I'd be willing to transfer this data over to mdn/data (by mapping it to the `appliesto` field) if there is interest.

captainbrosset:
That sounds good to me. Fetching all of the webref data for CSS and checking all of the possible appliesto cases should be quick to do and would give a good idea of the complexity involved.

captainbrosset:
[...]

PR #607, opened 2022-09-26
contain-intrinsic- size, width, height

hamishwillee:
This adds the data for `contain-intrinsic-size`, `contain-intrinsic-width` (will extend it to support `contain-intrinsic-height` once width is "approved".

Note, I have never done one of these before so sorry if things very wrong. This depends on #606


css/properties.json
```diff
@@ -4220,6 +4220,49 @@
...
+    "percentages": "no",
+    "groups": [
+      "CSS Containment"
+    ],
+    "initial": "none",
+    "appliesto": "elementsForWhichSizeContainmentCanApply",
+    "computed": "asSpecified",
+    "order": "perGrammar",
+    "status": "experimental",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/contain-intrinsic-width"
```
hamishwillee, line 4264:
```suggestion
    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/contain-intrinsic-size"
```

css/properties.json
```diff
@@ -4220,6 +4220,49 @@
...
+    "percentages": "no",
+    "groups": [
+      "CSS Containment"
+    ],
+    "initial": "none",
+    "appliesto": "elementsForWhichSizeContainmentCanApply",
+    "computed": "asSpecified",
+    "order": "perGrammar",
+    "status": "experimental",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/contain-intrinsic-size"
```
hamishwillee, line 4264:
```suggestion
    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/contain-intrinsic-width"
```

css/properties.json
```diff
@@ -4220,6 +4220,49 @@
...
+      "contain-intrinsic-width",
+      "contain-intrinsic-height"
+    ],
+    "appliesto": "allElements",
+    "computed": [
+      "contain-intrinsic-width",
+      "contain-intrinsic-height"
+    ],
+    "order": "perGrammar",
+    "status": "experimental"
```
hamishwillee, line 4248:
```suggestion
    "status": "experimental",
    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/contain-intrinsic-size"
```

css/properties.json
```diff
@@ -4220,6 +4220,50 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/contain"
   },
+  "contain-intrinsic-size": {
+    "syntax": "[ none | <length> | auto <length> ]{1,2}",
```
hamishwillee, line 4224: This is copied from spec: https://w3c.github.io/csswg-drafts/css-sizing-4/#propdef-contain-intrinsic-size

css/properties.json
```diff
@@ -4220,6 +4220,50 @@
...
+    "media": "visual",
+    "inherited": false,
+    "animationType": [
+      "contain-intrinsic-width",
+      "contain-intrinsic-height"
```
hamishwillee, line 4229: NOte, I just copied the pattern of another case and listed all the longhand properties for all of these because the spec said "see individual properties".

css/properties.json
```diff
@@ -4220,6 +4220,50 @@
...
+    "appliesto": "allElements",
+    "computed": [
+      "contain-intrinsic-width",
+      "contain-intrinsic-height"
+    ],
+    "order": "perGrammar",
+    "status": "experimental",
+    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/contain-intrinsic-size"
+  },
+  "contain-intrinsic-width": {
```
hamishwillee, line 4251: This is from https://w3c.github.io/csswg-drafts/css-sizing-4/#propdef-contain-intrinsic-width

I'll do https://w3c.github.io/csswg-drafts/css-sizing-4/#propdef-contain-intrinsic-size when this part is "approved".

css/properties.json
```diff
@@ -4220,6 +4220,50 @@
...
+    "syntax": "none | <length> | auto <length>",
+    "media": "visual",
```
hamishwillee, line 4253: Note sure what is right for this one.

css/properties.json
```diff
@@ -4220,6 +4220,50 @@
...
+    "inherited": false,
+    "animationType": "byComputedValueType",
+    "percentages": "no",
+    "groups": [
+      "CSS Containment"
+    ],
+    "initial": "none",
+    "appliesto": "elementsForWhichSizeContainmentCanApply",
+    "computed": "asSpecified",
```
hamishwillee, line 4262: Is this correct? Spec says "as specified, with [`<length>`](https://w3c.github.io/csswg-drafts/css-values-4/#length-value) values computed". 

The values for length (or in this case width are always specified). However if you were to use the `auto <length>` value it might be determined from a stored value.

hamishwillee:
NOTE, at least some of this stuff now appears to be coming from somewhere - is appearing in my local builds.

css/properties.json
```diff
@@ -4220,6 +4220,50 @@
...
+      "contain-intrinsic-height"
+    ],
+    "groups": [
+      "CSS Containment"
+    ],
+    "initial": [
+      "contain-intrinsic-width",
+      "contain-intrinsic-height"
+    ],
+    "appliesto": "allElements",
```
wbamberg, line 4242: I think this is wrong? Wouldn't this be the same as for the longhand:

> elements with [size containment](https://w3c.github.io/csswg-drafts/css-contain-2/#size-containment)

?

css/properties.json
```diff
@@ -4220,6 +4220,50 @@
...
+    "syntax": "none | <length> | auto <length>",
+    "media": "visual",
+    "inherited": false,
+    "animationType": "byComputedValueType",
+    "percentages": "no",
+    "groups": [
+      "CSS Containment"
+    ],
+    "initial": "none",
+    "appliesto": "elementsForWhichSizeContainmentCanApply",
```
wbamberg, line 4261: This is an enum which has to map to one of the values in https://github.com/mdn/data/blob/main/l10n/css.json and https://github.com/mdn/data/blob/fea9f55167435f34fb39912151e1cc4288e7afb7/css/properties.schema.json#L162. So if this is new (which it looks like it is) then you need to update those files too.

(yes, it's awful. I think this is where you get when you try to represent prose as data. See also https://github.com/mdn/data/issues/585 for some interesting discussion of this.)

css/properties.json
wbamberg: I don't know. Maybe you need a new `asSpecifiedWithLengthValuesComputed` enum?

css/properties.json
wbamberg: `width` is correct here isn't it?

wbamberg (changes requested):
I hope this is somewhat helpful! tbh I don't understand a lot of this and there seems to be a lot of cargo-culting going on.

I would **love** to start generating this from webref so we can stop maintaining this.

css/properties.json
hamishwillee: Yes I do. I mean if this is about right, I'll add docs for other properties. Might add those now so it can all merge together.

css/properties.json
```diff
@@ -4220,6 +4220,50 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/contain"
   },
+  "contain-intrinsic-size": {
+    "syntax": "[ none | <length> | auto <length> ]{1,2}",
+    "media": "visual",
```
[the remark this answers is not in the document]
hamishwillee: Thanks, will omit.
```suggestion
```

css/properties.json
hamishwillee:
```suggestion
```

css/properties.json
hamishwillee: That is horrible. But then so is this whole thing. I added this value as part of https://github.com/mdn/data/pull/606

css/properties.json
hamishwillee: I think you're right - add the value in #606 
```suggestion
    "computed": "asSpecifiedWithLengthValuesComputed",
```

css/properties.json
hamishwillee: Yes. Thanks
```suggestion
    "appliesto": "elementsForWhichSizeContainmentCanApply",
```

css/properties.json
hamishwillee: Done.

hamishwillee:
Thanks @wbamberg . I've made the suggested changes and added in the other properties, so this should be good to go. Note though that #607 needs to merge first because of the new enum values.

hamishwillee:
Arrggg. Media is actually required @wbamberg . I'll try add in when I'm back on Friday. Or can we remove it from the schema?

hamishwillee:
@wbamberg I added back the media since removing it causes failures. This can't merge yet though until https://github.com/mdn/data/pull/608 is merged.

Can you merge both please?

hamishwillee:
@teoli2003 Is this something you might be able to review?

teoli2003 (approved):

hamishwillee:
@teoli2003 Thanks. Can you merge it too. I have no rights here.

wbamberg (approved):
Sorry to be slow Hamish.

wbamberg:
Actually @hamishwillee we're still getting line errors about `@media`.

wbamberg:
I just filed https://github.com/mdn/data/pull/611 which ought to make `media` optional.

hamishwillee:
Thanks guys!

PR #610, opened 2022-10-10
Add `<function-token>` to CSS Syntaxes

ramiy:
Source: https://www.w3.org/TR/css-syntax-3/#function-token-diagram


lahmatiy:
`function-token` is not a production, it's a token and equals to itself like other `*-token`. Proposed PR is a mistake.

PR #611, opened 2022-10-12
Make media optional in schema

wbamberg:
This ought to make `media` optional in the CSS schema. If I have understood it properly 🤞 .

We have not rendered `media` for a long time on MDN (https://github.com/mdn/kumascript/pull/1356) and I think the specs have stopped including it as well. So it's silly for the linter to complain about it.


teoli2003 (approved):
👍

PR #616, opened 2022-10-22
border-image-repeat applied to ::first-letter removed

sunkanmii:
According to the official doc - [https://www.w3.org/TR/css-backgrounds-3/#border-image-repeat](https://www.w3.org/TR/css-backgrounds-3/#border-image-repeat), and testing I have done personally, the CSS border-image-repeat property doesn't apply just to the ::first-letter of the paragraph tag, it doesn't display any border image at all or even repeats what was placed on it.

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -2899,9 +2899,6 @@
     "appliesto": "allElementsExceptTableElementsWhenCollapse",
     "computed": "asSpecified",
     "order": "uniqueOrder",
-    "alsoAppliesTo": [
-      "::first-letter"
-    ],
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/border-image-repeat"
   },


wbamberg:
They do apply though, don't they? See https://codepen.io/willbamberg/pen/yLENRGG?editors=1100 for example.

> According to the official doc - https://www.w3.org/TR/css-backgrounds-3/#border-image-repeat

The spec doesn't list pseudo-elements in the property definition, but with the pseudo-element itself. https://w3c.github.io/csswg-drafts/css-pseudo/#first-letter-styling lists "border properties", which seems like it woud include `border-image-*` properties.

sunkanmii:
Ah. You're right actually. It wasn't working in my local environment, so I concluded it wasn't working.

sunkanmii:
> ...surrounding content is based on the [border-width](https://www.w3.org/TR/css-backgrounds-3/#propdef-border-width) and [border-style](https://www.w3.org/TR/css-backgrounds-3/#propdef-border-style) properties only.

After reviewing and reading through the docs again, And according to the doc (as stated above), I found out that border-image can't work without `border-style` and `border-width` set. In my code, I didn't set `border-style` which made the borders not have any display.

Closing all my PRs related to this. Thanks for pointing it out!

wbamberg:
You're welcome @sunkanmii , glad I could help!

PR #617, opened 2022-10-22
border-image-outset applied to ::first-letter removed

sunkanmii:
According to the official doc - https://www.w3.org/TR/css-backgrounds-3/#border-image-outset, and testing I have done personally, the CSS border-image-outset property doesn't apply to the ::first-letter of the paragraph tag, it doesn't display any border image at all or even creates a distance between the paragraph's border image and its border box.


sunkanmii:
Apparently, they do apply.
Check out Will's https://github.com/mdn/data/pull/616#issuecomment-1295450899 in PR https://github.com/mdn/data/pull/616

PR #618, opened 2022-10-22
border-image-width applied to ::first-letter removed

sunkanmii:
According to the official doc - https://www.w3.org/TR/css-backgrounds-3/#border-image-width, and testing I have done personally, the CSS border-image-width property doesn't apply to the ::first-letter of the paragraph tag, it doesn't display any border image at all or even creates a border image width.


sunkanmii:
Apparently, they do apply.
Check out Will's https://github.com/mdn/data/pull/616#issuecomment-1295450899 in PR https://github.com/mdn/data/pull/616

PR #620, opened 2022-10-22
border-image-source applied to ::first-letter removed

sunkanmii:
According to the official doc - https://www.w3.org/TR/css-backgrounds-3/#border-image-source, and testing I have done personally, the CSS border-image-source property doesn't apply to the ::first-letter of the paragraph tag, it doesn't display any border image at all. 

It only applies to: *All elements, except internal table elements when border-collapse is collapse*


sunkanmii:
Apparently, they do apply.
Check out Will's comment in PR #616

PR #622, opened 2022-10-27
fix: correct scroll-timeline-axis and mask-repeat initial value

ludofischer:
1. The current scroll-animations spec says the initial value is 'block'. https://w3c.github.io/csswg-drafts/scroll-animations/#scroll-timeline-axis. `none` does not look like a legal value to me.
2. The `mask-repeat` initial value is  `repeat`: https://drafts.fxtf.org/css-masking/#the-mask-repeat
    You can even check that on the MDN page that says the initial value is `no-repeat` by setting `mask-repeat` to `initial`.

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -8627,7 +8627,7 @@
     "groups": [
       "CSS Animations"
     ],
-    "initial": "none",
+    "initial": "block",
     "appliesto": "scrollContainers",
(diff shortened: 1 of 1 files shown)


wbamberg (approved):
👍 thank you @ludofischer ! Wish we could get on with populating these directly from the specifications.

PR #625, opened 2022-11-13
Fix typo in CSS name-repeat syntax definition

sideshowbarker:
Fixes https://github.com/mdn/content/issues/22262


wbamberg (approved):
👍 thank you @sideshowbarker ! I'll merge this because it looks like a good fix but it is incomplete because it looks like the translations are also wrong. So I will file a separate issue for that.

PR #629, opened 2022-12-05
Follow-up of #625 for `fr`

### LINKED ISSUE #627: Localized strings for `background-position-x` percentages are wrong
https://github.com/mdn/data/pull/625 fixed an issue where the "percentages" bit for `background-position-x` incorrectly used `height` instead of `width`. But it only addressed en-US, and AFAICT the translated versions are still incorrect.

cc @mdn/localization-team-leads - if you file PRs to address this please tag me for a review :)

SphinxKnight:
Should fix #627 for `fr`


SphinxKnight:
@wbamberg if you have time for this :) (low priority I'd think)

wbamberg (approved):
👍 thank you @SphinxKnight

PR #609, opened 2022-10-10
scroll-timeline-* properties

hamishwillee:
Added the named scroll timeline properties. Other docs work can be tracked in #21069

As previously, attempting to follow existing patterns.

PR #631, opened 2022-12-15
fix: scroll-timeline-axis and scroll-timeline-name are properties

pyoor:
https://github.com/mdn/data/pull/609 incorrectly recorded the `scroll-timeline` syntax as `<scroll-timeline-name> || <scroll-timeline-axis>`.  This is incorrect as `scroll-timeline-axis` and `scroll-timeline-name` are properties and need to be quoted as such.

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -8610,7 +8610,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/scroll-snap-type-y"
   },
   "scroll-timeline": {
-    "syntax": "<scroll-timeline-name> || <scroll-timeline-axis>",
+    "syntax": "<'scroll-timeline-name'> || <'scroll-timeline-axis'>",
     "media": "visual",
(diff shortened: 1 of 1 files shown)


lahmatiy:
According to latest state of [spec](https://www.w3.org/TR/scroll-animations-1), `scroll-timeline-name` and `scroll-timeline-axis` properties accept a list of values instead of a single one now. So their definitions need to be updated as well:

```
<scroll-timeline-name> = none | [<custom-ident>]#
<scroll-timeline-axis> = [ block | inline | vertical | horizontal ]#
```

teoli2003 (approved):

teoli2003:
Hi! In order to merge this, commits need to be signed: https://docs.github.com/articles/about-gpg/

bsmth:
>commits need to be signed

This should not be a blocker any more

teoli2003:
Thank you!

PR #635, opened 2023-01-04
Fix definitions for border-block and border-inline

wbamberg:
Fixes https://github.com/mdn/data/issues/634 (I think?)

-> https://w3c.github.io/csswg-drafts/css-logical/#propdef-border-block


teoli2003 (approved):
👍

PR #646, opened 2023-03-01
Quote wildcard token in view-transition syntax

pyoor:
### Description

According to the CSS value definition syntax, the wildcard token must be quoted unless used as a multiplier.  See the definition of the view-transition syntax [here](https://www.w3.org/TR/css-view-transitions-1/#typedef-pt-name-selector).


teoli2003 (approved):
This is correct. Thank you!

PR #647, opened 2023-03-01
Wrap view-transition arguments in an explicit group

pyoor:
### Description
I should have noticed this when I submitted https://github.com/mdn/data/pull/646, but CSS function notation expects a single argument.  In this case, it requires wrapping the arguments in an explicit group.

See also https://github.com/w3c/csswg-drafts/issues/7582.


teoli2003 (approved):
👍

PR #650, opened 2023-03-18
Add `relativeToWidthAndHeight` for `en-US`

yarusome:
### Description

This PR adds the `relativeToWidthAndHeight` l10n key for `en-US`.

### Motivation

This key is [used](https://github.com/yarusome/mdn-data/blob/main/css/properties.json#L6982) to show how the precentages are solved by the `offset-anchor` CSS property.

### Additional details

The definition of `offset-anchor` in Motion Path Module Level 1:
https://drafts.fxtf.org/motion/#offset-anchor-property


teoli2003 (approved):
👍

PR #652, opened 2023-03-18
Fix `appliesto` for `content-visibility`

yarusome:
### Description

This PR fixes the `appliesto` key for the `content-visibility` CSS property.

### Motivation

This key uses the non-existent `elementsForWhichLayoutContainmentCanApply` l10n key. Also, the spec shows that `content-visibility` applies to elements for which size, instead of layout, containment can apply.

### Additional details

The definition of `content-visibility` in CSS Containment Module Level 2:
https://w3c.github.io/csswg-drafts/css-contain/#content-visibility

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -4374,7 +4374,7 @@
       "CSS Containment"
     ],
     "initial": "visible",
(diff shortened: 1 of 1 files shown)


teoli2003 (approved):
👍

PR #655, opened 2023-03-23
Add  data for `font-variant-emoji`

yarusome:
### Description

This PR adds the data for the `font-variant-emoji` CSS property.

### Motivation

See https://github.com/mdn/content/issues/25441.

### Additional details

  1. Many properties in CSS Font Module lack a `media` field, but I followed the examples of other `font-variant-*` setting the `media` field to `visual`.
  2. The `alsoAppliesTo` field ***might not*** be correct due to the phrasing in the spec. See https://github.com/w3c/csswg-drafts/issues/8635.

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5036,6 +5036,26 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/font-variant-east-asian"
   },
+  "font-variant-emoji": {
+    "syntax": "normal | text | emoji | unicode",
+    "media": "visual",
+    "inherited": true,
+    "animationType": "discrete",
+    "percentages": "no",
+    "groups": [
+      "CSS Fonts"
+    ],
+    "initial": "normal",
+    "appliesto": "allElementsAndText",
+    "computed": "asSpecified",
+    "order": "perGrammar",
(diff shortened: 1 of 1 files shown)


yarusome:
A side question: It's a task on my backlog to add all the missing `zh-CN` l10n keys, but currently this repo has a deprecation note:

> **Note**
> We are in the process of deprecating the `mdn/data` package in favor of [`w3c/webref`](https://github.com/w3c/webref).
> If you depend on this project, let us know in our community [GitHub discussions](https://github.com/mdn/mdn-community/discussions/categories/platform).
> Thank you.

Does that mean all these l10n keys will be useless in the future? It would be a pity to see these localized keys wasted.

teoli2003:
In the long term, we want to retire mdn/data. When possible, we are using different sources on MDN. This is the case for the syntax where MDN uses directly w3c/webref nowadays.

There are other consumers of mdn/data, though.

For l10n, as far as I know, no other alternatives to mdn/data have l10n support, so mdn/data is not going away soon. Finding a solution for l10n (maybe by having the translations living elsewhere) is a problem. So I don't think the translations will be lost as they are unique. I hope this helps.

teoli2003 (approved):
👍

PR #658, opened 2023-03-29
Fix capitalization of fragment

teoli2003:
In the Kuma era and before, capitalization was important for fragments. Nowadays, internal fragments are always lowercase.

This is important for `#Interpolation`, which must be `#interpolation`.

This should fix mdn/content#25425

(Thanks, @estelle, for investigating the problem)


wbamberg (approved):
👍

PR #662, opened 2023-04-24
New data type: overflow

estelle:
adding `<overflow>` as a data type as 5 properties use the exact same group of keywords.

part of https://github.com/mdn/content/pull/26395


teoli2003 (approved):
👍

PR #674, opened 2023-07-09
The `initial-value` descriptor in an `@property` rule takes any value.

romainmenke:
### Description

The `initial-value` descriptor of `@property` rules is not limited to strings.  
It permits any value.

### Motivation

`<string>` is incorrect.

### Additional details

see : https://drafts.css-houdini.org/css-properties-values-api/#the-syntax-descriptor

N/A


css/at-rules.json
```diff
@@ -455,7 +455,7 @@
         "status": "experimental"
       },
       "initial-value": {
-        "syntax": "<string>",
+        "syntax": "<declaration-value>",
```
OnkarRuikar, line 458:
```suggestion
        "syntax": "<declaration-value>?",
```
https://drafts.css-houdini.org/css-properties-values-api/#initial-value-descriptor

teoli2003 (approved):
👍

PR #689, opened 2023-09-08
chore(ci): Add conventional commits guideline to the PR template

OnkarRuikar:
Contributors are not following conventional commits so the releases are not happening frequently. The PR updates PR template and CONTRIBUTING docs.


OnkarRuikar:
We can also create [a workflow to check PR titles](https://github.com/MorrisonCole/pr-lint-action) when a PR is created, edited, or reopened.

bsmth (approved):
Thanks @OnkarRuikar - (next up is adding a CODEOWNERS so we get the review auto-assigned 😉)

PR #688, opened 2023-09-07
overlay, starting-style, discrete animations

chrisdavidmills:
#### Summary

This PR adds multiple related features that go together to enable the creation of exit/entrance animations using simple CSS, or to put it another way, animating to/from `display: none`, or animating elements when they have just been added to the DOM.

Specifically, these are:

- [`@starting-style`](https://chromestatus.com/feature/4515377717968896): Chrome 117
- [The `overlay` property](https://chromestatus.com/feature/5138724910792704): Chrome 117
- [Animating to/from `display: none` and `content-visibility: hidden`](https://chromestatus.com/feature/5154958272364544): Chrome 116
- [The `transition-behavior` property](https://chromestatus.com/feature/5071230636392448): Chrome 117 (this was already added, but it is related, so I checked it)

See my [research document](https://docs.google.com/document/d/1esudzJfl1xpALiPDJ5cdISWGT4zvhZOsiixSriqc_sk/edit) for more details of what this project aims to add.

PR #690, opened 2023-09-11
fix: descriptions of discrete animation behavior between visible and hidden

chrisdavidmills:
### Description

In https://github.com/mdn/data/pull/688, part of the work was to add descriptions of the new specific animation behavior of certain "discrete" animated properties — `display`, `content-visibility`, and `overlay` — when animating from a hidden state or a visible state. However, on talking to some Chrome engineering folks about this, I realized that some of my descriptions were not quite right. 

This PR aims to fix those descriptions.

The description I was sent by the Chrome engineering folk is as follows: 

"The idea behind "p = 1" is that during transitions between certain values for certain properties which we consider invisible to visible or vice versa, the browser will automatically choose the "visible" option for the duration of the animation it creates.
For example, if you transition from "display:none" to "display:block", the element will be "display:block" for the entire animation created by the transition.
And likewise if you transition from "display:block" to "display:none", the element will be "display:block" for the entire animation created by the transition.
However, if you transition from "display:flex" to "display:block", the element's display value will switch halfway through the animation.
This behavior is applied to content-visibility:hidden, display:none, visibility:hidden, and overlay:none."


teoli2003 (approved):
👍

PR #694, opened 2023-10-24
fix(css): replace value 'invert' with 'auto' for outline-color property

OnkarRuikar:
- related content PR: https://github.com/mdn/content/pull/29798


teoli2003 (approved):
👍

PR #697, opened 2023-12-28
Remove abandoned value + fix renamed ones

teoli2003:
Several features have been renamed or abandoned over the years.

Because they are listed incorrectly here, they lead to broken links in CSS/Reference.

This PR removed them and renamed `<timing-function>` into the `<easing-function>` as used everywhere.

### DIFF
--- a/docs/updating_css_json.md
+++ b/docs/updating_css_json.md
@@ -318,7 +318,7 @@ For shorthand properties, several entries are a list of the longhand properties 
 ```json
 {
   "animation": {
-    "syntax": "&lt;single-animation-name&gt; || &lt;time&gt; || &lt;timing-function&gt; || &lt;time&gt; || &lt;single-animation-iteration-count&gt; || &lt;single-animation-direction&gt; || &lt;single-animation-fill-mode&gt; || &lt;single-animation-play-state&gt;",
+    "syntax": "&lt;single-animation-name&gt; || &lt;time&gt; || &lt;easing-function&gt; || &lt;time&gt; || &lt;single-animation-iteration-count&gt; || &lt;single-animation-direction&gt; || &lt;single-animation-fill-mode&gt; || &lt;single-animation-play-state&gt;",
     "media": "visual",
     "inherited": false,
     "animatable": "no",
--- a/docs/updating_css_json.md
+++ b/docs/updating_css_json.md
@@ -313,39 +313,40 @@ The syntaxes for the values `<family-name>` and `<generic-family>` are stored li
 
 ## Example for a CSS shorthand property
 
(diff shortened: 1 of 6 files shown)


docs/updating_css_json.md
```diff
@@ -318,7 +318,7 @@ For shorthand properties, several entries are a list of the longhand properties
 ```json
 {
   "animation": {
-    "syntax": "&lt;single-animation-name&gt; || &lt;time&gt; || &lt;timing-function&gt; || &lt;time&gt; || &lt;single-animation-iteration-count&gt; || &lt;single-animation-direction&gt; || &lt;single-animation-fill-mode&gt; || &lt;single-animation-play-state&gt;",
+    "syntax": "&lt;single-animation-name&gt; || &lt;time&gt; || &lt;easing-function&gt; || &lt;time&gt; || &lt;single-animation-iteration-count&gt; || &lt;single-animation-direction&gt; || &lt;single-animation-fill-mode&gt; || &lt;single-animation-play-state&gt;",
```
dipikabh, line 321: What do you think about updating this syntax to:

```suggestion
    "syntax": "&lt;single-animation&gt;#",
```

I realize that you were only updating the `timing-function` to the more current `easing-function` and I was doing a casual comparison with the syntax on the `animation` page when I noticed this. 

The [Formal syntax section ](https://developer.mozilla.org/en-US/docs/Web/CSS/animation#formal_syntax) and the [spec](https://drafts.csswg.org/css-animations/#animation) list the syntax as:

`animation = <single-animation>#`

The value for the `"syntax"` key above on L321 is an enumeration of `<single-animation>` values. The `syntaxes.json` file does have an entry for [`single-animation`](https://github.com/mdn/data/blob/03779ab09774edfdf76e1a909907b33012ddbdd6/css/syntaxes.json#L743). Perhaps that's the one that needs to be updated, for example, to remove `<'animation-duration'>` and add `<time>` among other things?

Also to note, in the `"syntax"` key on L321, `&lt;time&gt;` appears twice.

dipikabh (changes requested):
Thanks for these updates, @teoli2003. All changes look good. I only have one comment in the `updating_css_json.md` file about the `animation` syntax. LMK what you think.

docs/updating_css_json.md
teoli2003: I switched to use `border` as an example of a shorthand property, as it still has the classical structure of shorthands (animation or background allow for several animations or background and use the `#` syntax).

dipikabh: This is a much better example. Thanks!

dipikabh (approved):
Thanks for these updates!

PR #706, opened 2024-01-20
fix(css): Remove ::first-line from margin* props

ferdnyc:
### Description

Correct the `alsoAppliesTo` for the CSS `margin` and `margin-{top,bottom,left,right.trim}` properties.

### Motivation

The data claims that `margin*` properties apply to the `::first-line` pseudoelement, which is untrue. They apply to `::first-letter`, but not `::first-line`. As reported by @Akindin in mdn/content#31806.

Fixes mdn/content#31806

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6384,8 +6384,7 @@
     ],
     "order": "uniqueOrder",
     "alsoAppliesTo": [
-      "::first-letter",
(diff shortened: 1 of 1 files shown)


wbamberg (approved):
👍 thank you @ferdnyc !

PR #707, opened 2024-01-20
feat(css): alsoAppliesTo data for bg-pos-x/bg-pos-y

ferdnyc:
### Description

Copy the `alsoAppliesTo` data from `background-position` to `background-position-x` and `background-position-y`, an obvious derivation from the shorthand property.

### Motivation

Provide more complete/detailed "Formal definition" data for MDN CSS property reference.


teoli2003 (approved):
👍 Thank you!

Issue #713, opened 2024-02-05
mask-position Initial value


kimgaeun77:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/mask-position

### What information was incorrect, unhelpful, or incomplete?

The page says that the mask-position Initial value is center, but when I tested it in Chrome, Firefox, etc., it was not center.

### What did you expect to see?

I expected mask-position center as shown on the page.

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/css/mask-position`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/mask-position
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/css/mask-position/index.md
* Last commit: https://github.com/mdn/content/commit/c77cfcd17e85db6c1b93160c70668f2ff6c2809c
* Document last modified: 2023-07-20T07:38:29.000Z

</details>

bsmth:
This is now fixed in the linked PR. Thanks a lot for raising and fixing 👍🏻

kimgaeun77:
Yes, thank you too😊I have one question.
It doesn't seem to have changed in the mdn official document(https://developer.mozilla.org/en-US/docs/Web/CSS/mask-position) yet. Can you tell when it will change?

bsmth:
>I have one question.

Sure thing: your changes have been included in https://github.com/mdn/data/pull/715. This new version will be picked up automatically as a bumped dependency in Yari -> https://github.com/mdn/yari/blob/main/CHANGELOG.md and it should be visible when builds are done with this new version. It may take a few days, but there's no other manual tasks outstanding, as far as I know.

kimgaeun77:
Thank you so much for letting me know😊Have a nice day!!👍

PR #714, opened 2024-02-15
fix(css): replace value 'center' with '0% 0%' for mask-position inital value

kimgaeun77:
…l value

### Description

The page says that the mask-position Initial value is center, but when I tested it in Chrome, Firefox, etc., it was not center.
So I changed the initial value of the mask-position property from center to 0% 0%.

### Motivation

It will be helpful for people working with CSS by referring to the mdn document.

### Additional details

https://github.com/mdn/data/issues/713
https://developer.mozilla.org/en-US/docs/Web/CSS/mask-position

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6874,7 +6874,7 @@
     "groups": [
       "CSS Masking"
     ],
-    "initial": "center",
+    "initial": "0% 0%",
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Thank you!

(see https://www.w3.org/TR/css-masking-1/#the-mask-position)

PR #724, opened 2024-04-08
enhance(publish-release): inline workflow action + allow manual execution

caugner:
### Description

Inlines the publish-release action.

### Motivation

This change is in order to debug and resolve the publish errors.


caugner (commented):
Self-approval.

PR #728, opened 2024-04-11
Add CSS Anchor Positioning data

chrisdavidmills:
### Description

[CSS Anchor Positioning](https://www.w3.org/TR/css-anchor-position-1/) is set to be released in Chrome 125. This PR adds CSS data for the new properties, functions, and at-rules in the spec.

Note that I have only added minimal syntax data for the new features; I have also not updated the syntax for existing features that have new values as a result of this spec. This is because we no longer use the syntax fields on MDN. I am trying to add the minimum syntax possible to pass linting.


teoli2003 (approved):
👍

PR #748, opened 2024-07-22
fix: CSS zoom should be a not animatable property

yomotsu:
### Description

Fix the animatable type of CSS zoom to be not-animatable.

### Motivation

According to the [WD of the CSS zoom property](https://drafts.csswg.org/css-viewport/#zoom-property), the animation type of the zoom property is not animatable.
Some browsers, such as Firefox, which recently started supporting zoom, treat it as animatable though.

Here is a capture of the WD page.
image

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -10281,7 +10281,7 @@
     "syntax": "normal | reset | <number> | <percentage>",
     "media": "visual",
     "inherited": false,
-    "animationType": "integer",
+    "animationType": "notAnimatable",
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Good catch! Thank you :)

Browsers are all over the place with the zoom animation, but the data should be aligned with the spec. [I filed a bug for Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1917602), let’s see how it goes.

PR #759, opened 2024-08-20
fix: add mdn url for CSS offset-anchor

JayaKrishnaNamburu:
### Description

Adding `mdn_url` for `offset-anchor` property in the `properties.json`.

### Motivation

When consuming `properties.json` and finding the `mdn-_url` for the properties. I noticed that `mdn_url` for the `outline-anchor` property is missing.


bsmth (approved):
Great, thank you!

PR #761, opened 2024-09-05
feat(ci): Update depbot to check GH actions

bsmth:
There are warnings that actions used in this repo use deprecated Node versions. Adding depbot for `github-actions` ecosystem.

See https://github.com/mdn/content/blob/main/.github/dependabot.yml for reference


pepelsbey (approved):
Looks good! Thank you :)

PR #766, opened 2024-09-16
chore: update linear-gradient syntax to use optional color-interpolation-method

JayaKrishnaNamburu:
### Description

The syntax for `linear-gradient` is missing `color-interpolation-method` which had a baseline support from May 2023. You can get more details about the property here
https://developer.mozilla.org/en-US/docs/Web/CSS/color-interpolation-method

It is even missing from the formal syntax of linear-gradient property page in [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient#formal_syntax)

Here is the [CSS-Image-Module-4](https://drafts.csswg.org/css-images-4/#linear-gradients) with it's updated syntax for linear-gradients to support `color-interpolation-method`.

### Motivation

When trying to pass `linear-gradient` values using 3rd party property parsers like `css-tree`. It property is failing it is reading from mdn which had outdated syntax for linear-gradient. 
Here is an example
https://stackblitz.com/edit/vitejs-vite-147ich?file=main.js

### Additional details

CSS Image Module 4 -> https://drafts.csswg.org/css-images-4/#linear-gradients


bsmth (approved):
Super, looks great, thank you!

PR #769, opened 2024-10-06
feat: rename `inset-area` to `position-area`

skyclouds2001:
### Additional details

spec at https://drafts.csswg.org/css-anchor-position-1/#position-area

see also https://github.com/mdn/content/issues/35700, and the spec update https://github.com/w3c/csswg-drafts/pull/10592


pepelsbey (approved):
Looks good! Thank you :)

PR #773, opened 2024-10-12
feat: add data for `interpolate-size` property

skyclouds2001:
### Description

spec at https://drafts.csswg.org/css-values-5/#interpolate-size


pepelsbey (approved):
Looks good! Thank you :)

PR #774, opened 2024-10-14
feat: update `text-wrap` to be a shorthand property

Gusted:
### Description

Specification: https://drafts.csswg.org/css-text-4/#propdef-text-wrap.
Make `text-wrap` a shorthand property of `text-wrap-mode` and `text-wrap-style`.

### Motivation

Practically this will mean `text-wrap: auto` becomes valid according to the `properties.json` syntax.

### Additional details

N/A

N/A


pepelsbey (approved):
Looks good! Thank you :)

PR #775, opened 2024-10-18
chore(css): remove `column-fill: balance-all`

caugner:
### Description

Removes an value that has not been implemented by any browser.

See:
- https://github.com/mdn/content/pull/34561
- https://github.com/mdn/browser-compat-data/pull/23578
- https://github.com/mdn/interactive-examples/pull/2853


pepelsbey (approved):
Oh, good catch! Thank you :)

PR #783, opened 2024-11-18
feat(css): Add missing syntax for `stroke` property

skyclouds2001:
### Description

add a <paint> syntax and use it for both `stroke` and `fill` property

### Additional details

(`fill` and `stroke` takes the same value)

https://svgwg.org/svg2-draft/painting.html#SpecifyingPaint
https://svgwg.org/svg2-draft/painting.html#FillProperty
https://svgwg.org/svg2-draft/painting.html#StrokeProperty


pepelsbey (approved):
Looks good! Thank you :)

PR #784, opened 2024-11-18
feat(css): Update `<alpha-value>` usage

skyclouds2001:
### Description

for css properties, `<alpha-value>` has been replaced with `<opacity-value>`

see https://github.com/w3c/csswg-drafts/commit/d995b3485fb98ee7557449fd0ca7c3e79a11c8c3 for `opacity`, rest seems to be so when added

`opacity` https://drafts.csswg.org/css-color/#transparency
`fill-opacity` https://drafts.fxtf.org/fill-stroke/#fill-opacity
`shape-image-threshold` https://drafts.csswg.org/css-shapes/#shape-image-threshold-property

also, for css functions, it still keep to use `<alpha-value>`

suggest: merge https://github.com/mdn/data/pull/786 first

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -4766,7 +4766,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/fill"
   },
   "fill-opacity": {
-    "syntax": "<opacity-value>",
+    "syntax": "<'opacity'>",
     "media": "visual",
     "inherited": true,
     "animationType": "byComputedValueType",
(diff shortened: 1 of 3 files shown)


css/properties.json
```diff
@@ -4766,7 +4766,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/fill"
   },
   "fill-opacity": {
-    "syntax": "<alpha-value>",
+    "syntax": "<opacity-value>",
```
skyclouds2001, line 4769:
```suggestion
    "syntax": "<'opacity'>",
```

pepelsbey (approved):
Looks good! Thank you :)

PR #786, opened 2024-11-19
fix(css): fill-opacity should accept percentages

pamelalozano16:
### Description
`fill-opacity` should accept percentages (https://developer.mozilla.org/en-US/docs/Web/CSS/fill-opacity)


css/properties.json
```diff
@@ -4766,7 +4766,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/fill"
   },
   "fill-opacity": {
-    "syntax": "<alpha-value>",
+    "syntax": "<number> | <percentage>",
```
skyclouds2001, line 4769: maybe [<'opacity'>](https://developer.mozilla.org/en-US/docs/Web/CSS/opacity) or <opacity-value> is better?

see https://github.com/mdn/data/pull/784

css/properties.json
```diff
@@ -4766,7 +4766,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/fill"
   },
   "fill-opacity": {
-    "syntax": "<alpha-value>",
+    "syntax": "<opacity>",
```
skyclouds2001, line 4769:
```suggestion
    "syntax": "<'opacity'>",
```

note: `<opacity>` is not equal to `<'opacity'>`

https://github.com/mdn/data/blob/8ec6d5057d8b774c185e33fe684b5af4dc8f3333/css/properties.json#L9576-L9591

pamelalozano16: Thanks!

pepelsbey (approved):
Looks good! Thank you :)

PR #597, opened 2022-08-27
Add check for syntax parsing and connectivity, fix missed syntaxes

lahmatiy:
There are pretty common problems when updating CSS syntax dictionaries:
- added/updated definition can't be parsed;
- a new syntax has broken references to other definitions, i.e. a syntax definition it refers to is missed in dictionaries.

This PR adds new 2 types of examination:
- every syntax definition can be parsed using [CSS Units and Values ](https://www.w3.org/TR/css-values-4/#value-defs) rules (CSSTree is using as a parser);
- all parsed syntax definitions has no broken references (i.e. `<type>` and `<'property'>`), i.e. refer to existing definitions only.

The new checking rules found issues (there are connectivity issues only, because parse errors are ignored in some cases, see bellow):
```
Check CSS syntax connectivity...
  css/at-rules.json "@document" prelude used missed syntax definition <url>
  css/at-rules.json "@font-face/src" used missed syntax definition <url>
  css/at-rules.json "@font-face/unicode-range" used missed syntax definition <unicode-range>
  css/at-rules.json "@import" prelude used missed syntax definition <url>
  css/at-rules.json "@import" prelude used missed syntax definition <declaration>
  css/at-rules.json "@namespace" prelude used missed syntax definition <url>
  css/properties.json "-moz-binding" used missed syntax definition <url>
  css/properties.json "cursor" used missed syntax definition <url>
  css/properties.json "cursor" used missed syntax definition <x>
  css/properties.json "cursor" used missed syntax definition <y>
  css/properties.json "offset-path" used missed syntax definition <url>
  css/selectors.json ":has" used missed syntax definition <forgiving-relative-selector-list>
  css/selectors.json ":is" used missed syntax definition <forgiving-selector-list>
  css/selectors.json ":lang" used missed syntax definition <language-code>
  css/selectors.json "::cue" used missed syntax definition <selector>
  css/selectors.json "::cue-region" used missed syntax definition <selector>
  css/syntaxes.json "attr()" used missed syntax definition <attr-name>
  css/syntaxes.json "attr()" used missed syntax definition <attr-fallback>
  css/syntaxes.json "clip-source" used missed syntax definition <url>
  css/syntaxes.json "filter-function-list" used missed syntax definition <url>
  css/syntaxes.json "image" used missed syntax definition <url>
  css/syntaxes.json "image-src" used missed syntax definition <url>
  css/syntaxes.json "keyframe-block" used missed syntax definition <declaration-list>
[...]

PR #790, opened 2024-11-22
feat(css): Add missing `<coord-box>` `<offset-path>` syntax for `offset-path` property

skyclouds2001:
### Description

https://drafts.fxtf.org/motion/#propdef-offset-path

https://drafts.fxtf.org/motion/#typedef-offset-path

https://www.w3.org/TR/css-box-4/#typedef-coord-box
https://www.w3.org/TR/css-box-4/#typedef-paint-box

reported in https://github.com/mdn/data/pull/597

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -188,6 +188,9 @@
   "contrast()": {
     "syntax": "contrast( [ <number-percentage> ] )"
   },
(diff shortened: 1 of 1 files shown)


bsmth (approved):
LGTM, thanks 👍🏻

PR #797, opened 2024-11-23
fix(css): Mark `<shape>` type as obsolete

skyclouds2001:
### Additional details

### DIFF
--- a/css/types.schema.json
+++ b/css/types.schema.json
@@ -16,7 +16,8 @@
         "enum": [
           "standard",
           "nonstandard",
-          "experimental"
+          "experimental",
+          "obsolete"
         ]
       },
       "mdn_url": {
--- a/css/types.json
+++ b/css/types.json
@@ -245,7 +245,7 @@
     "groups": [
       "CSS Types"
     ],
-    "status": "standard",
+    "status": "obsolete",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/shape"
   },
   "string": {


css/types.schema.json
```diff
@@ -16,7 +16,8 @@
         "enum": [
           "standard",
           "nonstandard",
-          "experimental"
+          "experimental",
+          "obsolete"
```
bsmth, line 20: I believe we should be using `deprecated` instead of `obsolete`:

```suggestion
          "deprecated"
```

skyclouds2001: I choose "obsolete", not "deprecated" here because in other data, currently only "obsolete" is used, like:

https://github.com/mdn/data/blob/10594fb01e3949d4f74241fa01423014efba803e/css/functions.schema.json#L3-L10

https://github.com/mdn/data/blob/10594fb01e3949d4f74241fa01423014efba803e/css/properties.schema.json#L314-L321

It's OK to use "deprecated" here, but other "obsolete" usage should also change to "deprecated". This might cause a break change.

bsmth: I wonder where this terminology came from, maybe we should check the blame for these. If we're already using "obsolete", I've no objection to that, but we don't use "obsolete" in content or BCD as a technology status, so it would be nice to clarify.

skyclouds2001: I remember "obsolete" is used in content (I' m not sure how bcd goes) years ago, then later it is changed to "deprecated".

I guess simply the data repo is not following such changes.

skyclouds2001: some related discussions: https://github.com/orgs/mdn/discussions/566 and https://github.com/orgs/mdn/discussions/438
and mdn guide: https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Experimental_deprecated_obsolete

bsmth: >I guess simply the data repo is not following such changes.

Yes, I think this convention to use `deprecated` instead of `obsolete` hasn't propagated down to `mdn/data`. 

<details>

```bash
grep -r "status\"" ./ |sort  |uniq
.//css/at-rules.json:        "status": "experimental"
.//css/at-rules.json:        "status": "standard"
.//css/at-rules.json:    "status": "experimental",
.//css/at-rules.json:    "status": "nonstandard",
.//css/at-rules.json:    "status": "standard",
.//css/at-rules.md:      "status": "experimental"
.//css/at-rules.md:      "status": "standard"
.//css/at-rules.md:  "status": "standard"
.//css/at-rules.schema.json:            "status"
.//css/at-rules.schema.json:            "status": {
.//css/at-rules.schema.json:      "status"
.//css/at-rules.schema.json:      "status": {
.//css/functions.json:    "status": "experimental",
.//css/functions.json:    "status": "nonstandard"
.//css/functions.json:    "status": "nonstandard",
.//css/functions.json:    "status": "standard",
.//css/functions.schema.json:        "$ref": "#/definitions/status"
.//css/functions.schema.json:      "status"
.//css/functions.schema.json:      "status": {
.//css/functions.schema.json:    "status": {
.//css/properties.json:    "status": "experimental"
.//css/properties.json:    "status": "experimental",
.//css/properties.json:    "status": "nonstandard"
.//css/properties.json:    "status": "nonstandard",
.//css/properties.json:    "status": "obsolete",
.//css/properties.json:    "status": "standard"
.//css/properties.json:    "status": "standard",
.//css/properties.md:  "status": "standard"
.//css/properties.schema.json:        "$ref": "#/definitions/status"
.//css/properties.schema.json:      "status"
.//css/properties.schema.json:      "status": {
.//css/properties.schema.json:    "status": {
.//css/selectors.json:    "status": "experimental",
.//css/selectors.json:    "status": "nonstandard"
.//css/selectors.json:    "status": "nonstandard",
.//css/selectors.json:    "status": "standard",
.//css/selectors.md:  "status": "standard"
.//css/selectors.schema.json:      "status"
.//css/selectors.schema.json:      "status": {
.//css/types.json:    "status": "obsolete",
.//css/types.json:    "status": "standard"
.//css/types.json:    "status": "standard",
.//css/types.md:  "status": "standard"
.//css/types.schema.json:      "status"
.//css/types.schema.json:      "status": {
.//css/units.json:    "status": "standard"
.//css/units.md:  "status": "standard"
.//css/units.schema.json:      "status"
.//css/units.schema.json:      "status": {
.//docs/updating_css_json.md:    "status": "standard",
```

</details>

Let's stick with `obsolete` for now and look at renaming that in future

bsmth (approved):
MDN URL for reference: https://developer.mozilla.org/en-US/docs/Web/CSS/shape, the note mentions alternatives

PR #801, opened 2024-11-23
feat(css): Sync property status per bcd

skyclouds2001:
### Additional details


bsmth (approved):
All looking good again, nice 👍🏻

PR #802, opened 2024-11-26
fix(css): Fix syntax for white-space

skyclouds2001:
### Description

currently white-space only available as a shorthand property for white-space-collapse and text-wrap only, not for white-space-trim (which didn't support by any browser at present)

https://github.com/mdn/browser-compat-data/blob/c933819e955e470ff5d6948842f7dbda3a62fae1/css/properties/white-space.json#L309-L347

### Motivation

reported in https://github.com/mdn/data/pull/597

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -10590,7 +10590,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/visibility"
   },
   "white-space": {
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Looks good, I would link here for some more info for traceability

- [x] https://github.com/mdn/content/pull/26781

PR #803, opened 2024-11-26
feat(css): add mdn_url for features with documentations

skyclouds2001:
### Description

these features has its documentations, so it's fine to add a mdn_url key

### Additional details


bsmth (approved):
Nice, thanks. All look good to me :+1:

PR #805, opened 2024-11-28
feat(css): Allow at-rule descriptors also contain mdn_url key

skyclouds2001:


css/at-rules.json
```diff
@@ -123,7 +133,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@document"
   },
   "@font-palette-values": {
-    "syntax": "@font-palette-values <dashed-ident> { <declaration-list> }",
+    "syntax": "@font-palette-values <dashed-ident> {\n  <declaration-list>\n}",
```
bsmth, line 136: Is this intentional?

bsmth (commented):
One comment open about syntax changes, otherwise looking great, thanks.

css/at-rules.json
skyclouds2001: Yes, just a formatting adjustment to keep it consistent with other at-rules

bsmth (approved):
LGTM, thanks!

PR #809, opened 2024-12-01
fix(css): Fix syntax for `scroll-timeline-name`

skyclouds2001:
### Description

see https://drafts.csswg.org/scroll-animations/#propdef-scroll-timeline-name

### Motivation

check in chromium's devtool console

image


bsmth (approved):
Great, tnx

PR #810, opened 2024-12-01
feat(css): Support for `word-break:auto-phrase`

skyclouds2001:
### Description

the feature has shipped in chrome M119 for ja locale (in CSS Text 4)

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
https://drafts.csswg.org/css-text/#word-break-property
https://drafts.csswg.org/css-text-4/#word-break-property


pepelsbey (approved):
Looks good! Thank you :)

PR #813, opened 2024-12-02
feat(css): Update syntax for sizing properties

skyclouds2001:
### Description

sizing properties include `height` `max-height` `max-width` `min-height` `min-width` `width`

merge `<length> | <percentage>` to `<length-percentage [0,∞]>` per specification
update `fit-content(<length-percentage>)` to `fit-content(<length-percentage [0,∞]>)` per specification
add missing `<calc-size()>` function syntax, which is supported in chromium M129
add missing `<anchor-size()>` function syntax, which is supported in chromium M125
add non-standard `-webkit-fill-available` value, its standard value is `stretch`, which currently is not supported per BCD

### Additional details

https://drafts.csswg.org/css-sizing/#propdef-width
https://drafts.csswg.org/css-sizing/#propdef-height
https://drafts.csswg.org/css-sizing/#propdef-min-width
https://drafts.csswg.org/css-sizing/#propdef-min-height
https://drafts.csswg.org/css-sizing/#propdef-max-width
https://drafts.csswg.org/css-sizing/#propdef-max-height

https://drafts.csswg.org/css-sizing-4/#valdef-width-stretch
https://drafts.csswg.org/css-sizing/#funcdef-width-fit-content
https://drafts.csswg.org/css-anchor-position/#anchor-size-fn

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5980,7 +5980,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/hanging-punctuation"
   },
   "height": {
-    "syntax": "auto | <length> | <percentage> | min-content | max-content | fit-content | fit-content(<length-percentage>)",
+    "syntax": "auto | <length-percentage [0,∞]> | min-content | max-content | fit-content | fit-content(<length-percentage [0,∞]>) | <calc-size()> | <anchor-size()>",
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #815, opened 2024-12-02
feat(css): Update syntax for margin properties

skyclouds2001:
### Description

merge `<length> | <percentage>` to `<length-percentage>` per specification

### Additional details

https://drafts.csswg.org/css-box/#margin-properties
https://drafts.csswg.org/css-logical/#margin-properties


pepelsbey (approved):
Looks good! Thank you :)

PR #816, opened 2024-12-02
feat(css): Update padding properties

skyclouds2001:
### Description

merge `<length> | <percentage>` to `<length-percentage [0,∞]>` per specification

### Additional details

https://drafts.csswg.org/css-logical/#padding-properties
https://drafts.csswg.org/css-box-4/#paddings


pepelsbey (approved):
Looks good! Thank you :)

PR #818, opened 2024-12-03
feat(css): remove non-supported caption-side values

skyclouds2001:
### Description

per bcd and mdn doc, these logical values havn't been supported by any platform

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/caption-side


pepelsbey (approved):
Oh, good catch! Thank you :)

PR #821, opened 2024-12-09
style(css,l10n): sort json files

skyclouds2001:
### Description

sort json data by alphabetical order

using https://github.com/skyclouds2001/mdn-tools/blob/master/css-json-sort.js


pepelsbey (approved):
Looks good! Thank you :)

PR #822, opened 2024-12-09
fix(l10n): fix incorrect punctuation marks

skyclouds2001:


pepelsbey (approved):
Looks good! Thank you :)

Issue #819, opened 2024-12-05
Missed syntaxes in 2.13.0


lahmatiy:
### What information was incorrect, unhelpful, or incomplete?

The issue of missing syntaxes in the new version of `mdn/data` was observed during an attempt to update `mdn/data` in `css-tree`. This was identified by the output of the improved linter from the rejected PR #597 as well:
```
Check CSS syntax connectivity...
  css/properties.json "font-palette" used missed syntax definition <palette-mix()>
  css/properties.json "height" used missed syntax definition <calc-size()>
  css/properties.json "height" used missed syntax definition <anchor-size()>
  css/properties.json "max-height" used missed syntax definition <calc-size()>
  css/properties.json "max-height" used missed syntax definition <anchor-size()>
  css/properties.json "max-width" used missed syntax definition <calc-size()>
  css/properties.json "max-width" used missed syntax definition <anchor-size()>
  css/properties.json "min-height" used missed syntax definition <calc-size()>
  css/properties.json "min-height" used missed syntax definition <anchor-size()>
  css/properties.json "min-width" used missed syntax definition <calc-size()>
  css/properties.json "min-width" used missed syntax definition <anchor-size()>
  css/properties.json "width" used missed syntax definition <calc-size()>
  css/properties.json "width" used missed syntax definition <anchor-size()>
  css/syntaxes.json "color" used missed syntax definition <oklab()>
  css/syntaxes.json "color" used missed syntax definition <oklch()>
  css/syntaxes.json "color" used missed syntax definition <color()>
```

### What did you expect to see?

All syntaxes are defined in dictionaries

Issue #811, opened 2024-12-02
Repetitive CSS function data between functions.json and syntaxes.json


skyclouds2001:
### What information was incorrect, unhelpful, or incomplete?

many CSS function data are repetitive represented the functions.json file and syntaxes.json file, this may cause inconsistent and hard to under

for example:

https://github.com/mdn/data/blob/d572bcb6dd50f3be7acc87d6203c7b740a280cb2/css/functions.json#L18-L25

and

https://github.com/mdn/data/blob/d572bcb6dd50f3be7acc87d6203c7b740a280cb2/css/syntaxes.json#L44-L46

### What did you expect to see?

only keep CSS function data in functions.json

### Do you have anything more you want to share?

note the change may cause break change though

maybe should listen to other's opinions

bsmth:
I'm curious why we have this duplication; it would be good to investigate if it's possible to mirror or reference another member in future

PR #824, opened 2024-12-10
feat(css): add missing css functions to syntaxes

skyclouds2001:
### Description

reported in https://github.com/mdn/data/issues/819

current some css functions are in functions.json, but more css functions are in syntaxes.json, see https://github.com/skyclouds2001/mdn-tools/blob/master/results/missing_function.json for checking results

this PR add css functions that are in functions.json but not in syntaxes.json to syntaxes.json, which seems to be the current behavior

see also https://github.com/mdn/data/issues/811

note data is not check with spec but just simply copy from functions.json

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -11,8 +11,14 @@
   "alpha-value": {
     "syntax": "<number> | <percentage>"
   },
+  "anchor()": {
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #825, opened 2024-12-11
fix(css): fix animate type of `scrollbar-width`

skyclouds2001:
### Description

note that `scrollbar-width` only accepts three keywords value and can only switch between them

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width
https://drafts.csswg.org/css-scrollbars/#scrollbar-width


pepelsbey (approved):
Good catch! Thank you :)

PR #826, opened 2024-12-12
feat(css): Add data for `text-box-*`

skyclouds2001:
### Description

these properties are supported in chrome M133 and Safari M18.2

### Additional details

https://drafts.csswg.org/css-inline/#typedef-text-edge
https://drafts.csswg.org/css-inline/#propdef-text-box
https://drafts.csswg.org/css-inline/#propdef-text-box-trim
https://drafts.csswg.org/css-inline/#propdef-text-box-edge


pepelsbey (approved):
Looks good! Thank you :)

PR #827, opened 2024-12-12
feat(css): add missing css functions

skyclouds2001:
### Description

his PR add css functions that are in syntaxes.json but not in functions.json to functions.json, which seems to be the current behavior

see also https://github.com/mdn/data/issues/811

note data is not check with spec but just simply copy from syntaxes.json

this is part of missing ones, the rest will be emitted in another [PR](https://github.com/mdn/data/pull/829)


pepelsbey (approved):
Good catch! Thank you :)

PR #830, opened 2024-12-14
fix(css): fix group for `clamp()` css function

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/clamp
https://drafts.csswg.org/css-values/#funcdef-clamp


pepelsbey (approved):
Oh, good catch! Thank you :)

PR #831, opened 2024-12-16
fix(css): correct syntax for scale related css functions

skyclouds2001:
### Description

note that only those data in functions.json are not correct, those in syntaxes.json are correct

### Motivation

found with https://github.com/skyclouds2001/mdn-tools/blob/master/checks/function-consistent-check.js and reported in https://github.com/skyclouds2001/mdn-tools/blob/master/results/inconsistent_function.json

https://drafts.csswg.org/css-transforms-2/#funcdef-scale
https://drafts.csswg.org/css-transforms-2/#funcdef-scalex
https://drafts.csswg.org/css-transforms-2/#funcdef-scaley
https://drafts.csswg.org/css-transforms-2/#funcdef-scale3d
https://drafts.csswg.org/css-transforms-2/#funcdef-scalez

### DIFF
--- a/css/functions.json
+++ b/css/functions.json
@@ -450,7 +450,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/filter-function/saturate"
   },
   "scale()": {
-    "syntax": "scale( <number> , <number>? )",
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #832, opened 2024-12-18
feat(css): Remove never-supported obsolute azimuth feature

skyclouds2001:
### Description

the feature is never supported by any platform, according to the [bcd](https://github.com/mdn/browser-compat-data/pull/2744), and it is already gone in bcd via https://github.com/mdn/browser-compat-data/pull/6945, it is also not in latest [spec](https://drafts.csswg.org/css-speech/) and it marked as obsolute


pepelsbey (approved):
Good catch! Thank you :)

PR #834, opened 2024-12-18
feat(css): Update syntax for `image-rendering` property `smooth` value

skyclouds2001:
### Description

the `image-rendering` property also support a `smooth` value, per bcd, it is supported by firefox

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering
https://drafts.csswg.org/css-images/#propdef-image-rendering


pepelsbey (approved):
Good catch! Thank you :)

PR #744, opened 2024-06-28
feat(css): Add text keyword to background-clip

TrySound:
### Description

Added `text` keyword to `background-clip` value.

### Motivation

All browsers already support unprefixed background-clip with text keyword.

### Additional details

- https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip

### Related issues and pull requests

- https://github.com/mdn/data/pull/201

PR #835, opened 2024-12-18
feat(css): Update syntax for `background-clip`

skyclouds2001:
### Description

update syntax for this property to be align with latest spec

in [stable version](https://drafts.csswg.org/css-backgrounds/#propdef-background-clip), its syntax is defined as `<visual-box>#`, whcih defines support for `border-box`, `padding-box` and `content-box` values
in [latest version](https://drafts.csswg.org/css-backgrounds-4/#propdef-background-clip), its syntax is defined as `<bg-clip>#`, which additional defines support for `border-area` and `text` values

per bcd, `border-area` value is supported by Safari and `text` is supported in all major platform at present

see also https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip and related issue at csstree  https://github.com/csstree/csstree/issues/190

Takes place of https://github.com/mdn/data/pull/744 as the original one seems not to be working anymore

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -2240,7 +2240,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/background-blend-mode"
   },
   "background-clip": {
-    "syntax": "<box>#",
+    "syntax": "<bg-clip>#",
     "media": "visual",
     "inherited": false,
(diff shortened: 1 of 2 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #836, opened 2024-12-19
fix(css): Fix mdn_url for selectors and types

skyclouds2001:
### Description

fix mdn_url for selectors and types

for *Adjacent sibling combinator* and *Subsequent-sibling combinator*, rename and fix mdn_url to match the mdn/content

for IE-specifed feature, simply remove the mdn_url key

for non-existed mdn docs, remove the mdn_url key

for redirected mdn docs, update the mdn_url key

found using https://github.com/skyclouds2001/mdn-tools/blob/master/checks/mdn_url-check.js and result store in https://github.com/skyclouds2001/mdn-tools/blob/master/results/incorrect_mdn_url.json

### DIFF
--- a/css/selectors.json
+++ b/css/selectors.json
@@ -53,23 +53,23 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/Selector_list"
   },
-  "Adjacent sibling combinator": {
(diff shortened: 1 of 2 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #838, opened 2024-12-19
feat(css): Update synatx for `background-origin`

skyclouds2001:
### Description

spec has updated the syntax in https://github.com/w3c/csswg-drafts/commit/7dc439c83df8bd34885f74689f8cbc7dff77b5e0, which replace replaces `<box>` definition with `<visual-box>` reference. note the change does not affect the property's definiton.

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/background-origin
https://drafts.csswg.org/css-backgrounds/#propdef-background-origin

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -2303,7 +2303,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/background-image"
   },
   "background-origin": {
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #840, opened 2024-12-19
fix(css): Fix mdn_url for functions

skyclouds2001:
### Description

for redirected mdn docs, update the mdn_url key

found using https://github.com/skyclouds2001/mdn-tools/blob/master/checks/mdn_url-check.js and result store in https://github.com/skyclouds2001/mdn-tools/blob/master/results/incorrect_mdn_url.json

### DIFF
--- a/css/functions.json
+++ b/css/functions.json
@@ -244,7 +244,7 @@
       "CSS Color"
     ],
     "status": "nonstandard",
-    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/hsla"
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

bsmth:
@skyclouds2001 - not only relevant to this PR, but this is the first one I found with a link to your tool. Would you like to add it to https://github.com/mdn/awesome? There's a section dedicated to collecting tools like these and we'd like to highlight them so people don't recreate similar things independently.

skyclouds2001:
Sure! I'd love to.

PR #844, opened 2024-12-21
fix(css): incorrect mdn_url for `repeating-radial-gradient()`

skyclouds2001:
### Description

found using https://github.com/skyclouds2001/mdn-tools/blob/master/checks/mdn_url-check.js and report in https://github.com/skyclouds2001/mdn-tools/blob/master/results/incorrect_mdn_url.json

### DIFF
--- a/css/functions.json
+++ b/css/functions.json
@@ -540,7 +540,7 @@
       "CSS Color"
     ],
     "status": "standard",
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Thank you!

PR #845, opened 2024-12-22
style(css): update feature order

skyclouds2001:
### Description

found using https://github.com/skyclouds2001/mdn-tools/blob/master/checks/data-order-check.js and report in https://github.com/skyclouds2001/mdn-tools/blob/master/results/unordered_data.json


pepelsbey (approved):
Looks good! Thank you :)

PR #846, opened 2024-12-22
fix(css): fix status for custom properties

skyclouds2001:
### Description

per https://developer.mozilla.org/en-US/docs/Web/CSS/--*, its status should be standard


pepelsbey (approved):
Looks good! Thank you :)

PR #847, opened 2024-12-22
feat(css): remove mdn_url for all IE features

skyclouds2001:
### Description

their docs are removed from mdn docs long ago

found using https://github.com/skyclouds2001/mdn-tools/blob/master/checks/mdn_url-check.js
report in https://github.com/skyclouds2001/mdn-tools/blob/master/results/incorrect_mdn_url.json


pepelsbey (approved):
Looks good! Thank you :)

PR #851, opened 2024-12-23
feat(css): Add data for css `@view-transition` at-rule

skyclouds2001:
### Description

the data is in mdn web docs and bcd, but not in data yet

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/@view-transition
https://drafts.csswg.org/css-view-transitions-2/#at-view-transition-rule


bsmth (approved):
LGTM, thanks 👍🏻

Issue #812, opened 2024-12-02
Migrate ** to **


skyclouds2001:
### What information was incorrect, unhelpful, or incomplete?

see https://github.com/mdn/data/pull/797#discussion_r1860535126

mdn/content and mdn/browser-compat-data has already no longer use obsolete anymore, and replace it with deprecated

### What did you expect to see?

same as above

### Do you have any supporting links, references, or citations?

https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Experimental_deprecated_obsolete

bsmth:
In general, I think this is a good idea, it might be nice to check if this is a breaking change or not. I don't know of any consumers who rely on this status for anything.

Also for historical reasons:

- https://github.com/mdn/browser-compat-data/issues/265

PR #852, opened 2024-12-23
docs(css): Add obsolete value to status field for schema and docs

skyclouds2001:
### Description

this helps in preparing for https://github.com/mdn/data/issues/812


bsmth (approved):
That all makes sense to me, I'm adding a +1 for now, but it would be good to know if any consumers use a hardcoded status for anything (i.e., is this a breaking change for anything downstream?).

skyclouds2001:
I think it is OK currently for merging this PR, as this only make some additions by adding the *obsolete* field to those docs that does not have this field before, for consistent.

bsmth:
> this only make some additions by adding the _obsolete_ field

I agree, let's merge

PR #853, opened 2024-12-24
feat: add `@container` at-rule

Netail:
### Description

Add `@container` at rule

### Motivation

VSCode not correctly syntax hightlighting

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/@container
https://drafts.csswg.org/css-conditional-5/#container-rule


css/at-rules.json
```diff
@@ -120,6 +120,20 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@counter-style"
   },
+  "@container": {
+    "syntax": "@container <container-condition># {\n  <block-contents>\n}",
```
skyclouds2001, line 124:
```suggestion
    "syntax": "@container <container-condition># {\n  <block-contents>\n}",
```

the syntax need to add to syntaxes.json if it is not existed in that file

css/at-rules.json
```diff
@@ -120,6 +120,20 @@
...
+    "interfaces": [
+      "CSSGroupingRule",
+      "CSSConditionRule",
+      "CSSContainerRule"
+    ],
+    "groups": [
+      "CSS Conditional Rules",
+      "CSS Containment"
+    ],
```
skyclouds2001, line 133:
```suggestion
    "groups": [
      "CSS Conditional Rules"
    ],
```

since it is only defined in [CSS Conditional Rules](https://drafts.csswg.org/css-conditional/), not in [CSS Containment](https://drafts.csswg.org/css-contain/)

only "CSS Conditional Rules" should be presented

Netail:
Applied your feedback :)

css/at-rules.json
```diff
@@ -120,6 +120,20 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@counter-style"
   },
+  "@container": {
+    "syntax": "@container <container-condition># {\n  <block-contents>\n}",
+    "interfaces": [
+      "CSSGroupingRule",
+      "CSSConditionRule",
+      "CSSContainerRule"
+    ],
```
skyclouds2001, line 129: Currently sone at-rule includes its interface and inherited interface, while most are not. This may need desicion, but it OK to not modify this at present.

In docs:

> * `interfaces` (array of strings): These are the [CSSOM](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model) interfaces that belong to the at-rule.

For this reason I think only `CSSContainerRule` should be represented

css/at-rules.json
skyclouds2001: > * `groups` (array of strings): CSS is organized in modules like "CSS Fonts" or "CSS Animations". MDN organizes features in these groups as well — `groups` should contain the name of the module(s) the at-rule is defined in.

The docs for a reference.

css/at-rules.json
Netail: Mhm, I just followed some of the other at-rules which included the top 2 interfaces as well as their own

skyclouds2001: yes, it is true, so I mean what do you think of this case? If not sure, you could simply keep it.

Netail: I don't mind, let's keep it for now I guess

skyclouds2001: I think twice, and I think we should not keep inherit interfaces.

i. currently only three cases list inherit interfaces (including one that no list direct interface), while nine cases only list direct interface
ii. the inherit info is a API feature, which is better to maintain at [inheritance.json](https://github.com/mdn/data/blob/main/api/inheritance.json)

css/at-rules.json
bsmth: I think this is still outstanding, right? I don't get any matches on this branch for `@container` aside from ./css/at-rules.json:

```bash
grep -r "@container" ./
./css/at-rules.json:  "@container": {
./css/at-rules.json:    "syntax": "@container <container-condition># {\n  <group-rule-body>\n}",
./css/at-rules.json:    "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@container"
```

bsmth:
Thanks @Netail - there's a comment from Skyclouds above about adding to syntaxes.json, too: https://github.com/mdn/data/pull/853#discussion_r1896859569 

Do you want to take a look? Thank you!

css/at-rules.json
Netail: Added, but unsure about the following 3; https://github.com/mdn/data/pull/853/files#diff-bf23b9edf4e332a630c103de7fc52e4eecc7d5262a21ac5feb2cc04a76cf2f40R230-R238

css/syntaxes.json
skyclouds2001, line 1: it is better to make those syntaxes by alphabet order

css/syntaxes.json
```diff
@@ -206,6 +206,39 @@
...
+  },
+  "scroll-state-in-parens": {
+    "syntax": "( <scroll-state-query> ) | ( <scroll-state-feature> ) | <general-enclosed>"
+  },
+  "scroll-state-feature": {
+    "syntax": "<media-feature>"
+  },
+  "size-feature": {
+    "syntax": "<media-feature>"
+  },
```
skyclouds2001, line 235:
```suggestion
  "scroll-state-feature": {
    "syntax": "<media-query-list>"
  },
  "size-feature": {
    "syntax": "<media-query-list>"
  },
```

[`@media`](https://drafts.csswg.org/css-conditional-3/#at-ruledef-media) use [`<media-query-list>`](https://drafts.csswg.org/mediaqueries-5/#typedef-media-query-list), so I think we should use this, too

css/syntaxes.json
```diff
@@ -206,6 +206,39 @@
...
+  "style-feature": {
+    "syntax": "<declaration>: <declaration-value> | <custom-property-name>: <declaration-value>"
+  },
```
skyclouds2001, line 238: [`@import`](https://drafts.csswg.org/css-cascade-5/#at-ruledef-import) and [`@support`](https://drafts.csswg.org/css-conditional-3/#at-ruledef-supports) also use `<declaration>` to represent a single css declaration, so I think `<declaration>` has covered `<declaration-value>`

`<declaration> | <custom-property-name>: <declaration-value>`

skyclouds2001, line 238:
```suggestion
  "style-feature": {
    "syntax": "<declaration>"
  },
```

[`@import`](https://drafts.csswg.org/css-cascade-5/#at-ruledef-import) and [`@support`](https://drafts.csswg.org/css-conditional-3/#at-ruledef-supports) also use `<declaration>` to represent a single css declaration, so I think `<declaration>` has covered `<declaration-value>` yet

skyclouds2001 (commented):
honestly I'm not quite sure either...

css/syntaxes.json
Netail: Also the custom-property?

skyclouds2001: I think `<declaration>` has covered custom-property, but I'm not sure

css/at-rules.json
```diff
@@ -120,6 +120,17 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@counter-style"
   },
+  "@container": {
+    "syntax": "@container <container-condition># {\n  <group-rule-body>\n}",
```
bsmth, line 124: I think this should be `<block-contents>`, also suggested here: https://github.com/mdn/data/pull/853#discussion_r1896859569

```suggestion
    "syntax": "@container <container-condition># {\n  <block-contents>\n}",
```

bsmth (approved):
This is a +1 from me, I'll see if someone else can have another final check before merging.

Netail:
> This is a +1 from me, I'll see if someone else can have another final check before merging.

Found someone by any chance 😅

skyclouds2001:
I would say it is OK enough, if possible

bsmth:
I think we can merge this now, thanks a lot for the additions and your patience!

PR #854, opened 2024-12-24
fix(css): Fix group field for CSS Containment module

skyclouds2001:
### Description

these features are defined in [CSS Conditional Rules](https://drafts.csswg.org/css-conditional/), not in [CSS Containment](https://drafts.csswg.org/css-contain/)


bsmth (approved):
Nice, thank you!

PR #855, opened 2024-12-24
fix(css): Add group field for CSS Box Sizing module

skyclouds2001:
### Description

see spec [CSS Box Sizing](https://drafts.csswg.org/css-sizing/)


bsmth (approved):
Thanks!

PR #856, opened 2024-12-25
feat(css): Add/Fix interfaces field of at-rules

skyclouds2001:
### Additional details

see also discussion about _interfaces_ field at https://github.com/mdn/data/pull/853#discussion_r1896891060

### DIFF
--- a/css/at-rules.md
+++ b/css/at-rules.md
@@ -27,7 +27,7 @@ The 3 properties seen above are all required:
 
 There are 3 more properties that are optional:
 * `mdn_url` (string): a URL linking to the at-rule's page on MDN. This URL must omit the localization part of the URL (such as `en-US/`).
-* `interfaces` (array of strings): These are the [CSSOM](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model) interfaces that belong to the at-rule.
+* `interfaces` (array of strings): These are the [CSSOM](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model) interfaces that belong to the at-rule. Note that inherit interfaces like `CSSGroupingRule` or `CSSConditionRule` should not be included.
 * `descriptors` (object): see below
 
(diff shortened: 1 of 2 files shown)


bsmth (approved):
Makes sense, thanks a lot. Maybe we need a docs change to accompany this for [clarification on inheritance](https://github.com/mdn/data/blob/397906caf12d19f733d3e8ce5f087fac8e9eadac/css/at-rules.md?plain=1#L30).

skyclouds2001:
> Makes sense, thanks a lot. Maybe we need a docs change to accompany this for [clarification on inheritance](https://github.com/mdn/data/blob/397906caf12d19f733d3e8ce5f087fac8e9eadac/css/at-rules.md?plain=1#L30).

Add a sentence indicating that inherit interfaces should not be included.

css/at-rules.md
```diff
@@ -27,7 +27,7 @@ The 3 properties seen above are all required:
 
 There are 3 more properties that are optional:
 * `mdn_url` (string): a URL linking to the at-rule's page on MDN. This URL must omit the localization part of the URL (such as `en-US/`).
-* `interfaces` (array of strings): These are the [CSSOM](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model) interfaces that belong to the at-rule.
+* `interfaces` (array of strings): These are the [CSSOM](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model) interfaces that belong to the at-rule. Note that inherit interfaces like `CSSGroupingRule` or `CSSConditionRule` should not be included.
```
bsmth, line 30:
```suggestion
* `interfaces` (array of strings): These are the [CSSOM](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model) interfaces that belong to the at-rule. Note that inherited interfaces like `CSSGroupingRule` or `CSSConditionRule` should not be included.
```

bsmth:
Thank you!

PR #858, opened 2024-12-26
feat(css): Add missing css `color()` function

skyclouds2001:
### Description

this feature is supported in all major platform per bcd

some test log in chromium:

```log
CSS.supports('color','color(from hsl(0 100% 50%) xyz calc(x - 0.3) calc(y + 0.3) calc(z + 0.3) / calc(alpha - 0.1))')
true
CSS.supports('color','color(from hsl(0 100% 50%) srgb calc(r - 0.4) calc(g + 0.1) calc(b + 0.6) / calc(alpha - 0.1))')
true
CSS.supports('color','color(from hsl(0 100% 50%) srgb calc(r - 0.4) calc(g + 0.1) calc(b + 0.6) / none)')
true
CSS.supports('color','color(display-p3 0.3804 0.9921 0.1412)')
true
```

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color
https://drafts.csswg.org/css-color-5/#funcdef-color
https://drafts.csswg.org/css-color/#funcdef-color

### Related issues and pull requests

reported in https://github.com/mdn/data/issues/819

since this is the last case that not done in https://github.com/mdn/data/issues/819, so closes https://github.com/mdn/data/issues/819

PR #859, opened 2024-12-26
feat(css): Update/Extract <color> syntax

skyclouds2001:
### Description

update the syntax to match the specification

* extract `<color>` into `<color>` `<color-base>` `<color-function>` per spec
* add the missing `light-dark()` and `color-mix()` syntax usage （as `contrast-color()` `device-cmyk()` is never supported at present, it is better not to add it at present）
* remove values of `<deprecated-system-color>` that are already in `<system-color>`
* remove `transparent` value of `<named-color>`, as the spec regard it as a special keyword

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value
https://developer.mozilla.org/en-US/docs/Web/CSS/named-color
https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color
https://developer.mozilla.org/en-US/docs/Web/CSS/system-color
https://drafts.csswg.org/css-color-5/#typedef-color

Relates to https://github.com/csstree/validator/issues/29
Relates to https://github.com/mdn/data/issues/819

**Depends on:** 
- [x] https://github.com/mdn/data/pull/857
- [x] https://github.com/mdn/data/pull/858

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -144,7 +144,13 @@
     "syntax": "<url>"
   },
   "color": {
-    "syntax": "<rgb()> | <rgba()> | <hsl()> | <hsla()> | <hwb()> | <lab()> | <lch()> | <oklab()> | <oklch()> | <color()> | <hex-color> | <named-color> | <system-color> | <deprecated-system-color> | currentcolor | transparent"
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Nice, lgtm, thanks 👍🏻

PR #770, opened 2024-10-06
feat: update syntax for Anchor Positioning related properties

skyclouds2001:
### Description

https://drafts.csswg.org/css-anchor-position-1/#anchor-size-fn for `anchor-size()`
https://drafts.csswg.org/css-anchor-position-1/#position-anchor for `position-anchor`

PR #861, opened 2024-12-27
feat(css): add missing `anchor-size` syntax for `anchor-size()`

skyclouds2001:
### Description

this is forgotten to add in https://github.com/mdn/data/pull/770

### Motivation

```log
CSS.supports('height','anchor-size(block)')
true
CSS.supports('height','anchor-size(inline)')
true
CSS.supports('height','anchor-size(height)')
true
CSS.supports('height','anchor-size(width)')
true
CSS.supports('height','anchor-size(self-block)')
true
CSS.supports('height','anchor-size(self-inline)')
true
```

### Additional details

spec at https://drafts.csswg.org/css-anchor-position-1/#funcdef-anchor-size


bsmth (approved):
Nice 👍🏻

PR #863, opened 2024-12-27
feat(css): Remove lack-support, deprecated `input-security`

skyclouds2001:
### Description

data added in https://github.com/mdn/data/pull/512

found using https://github.com/skyclouds2001/mdn-tools/blob/master/checks/mdn_url-check.js and reported in https://github.com/skyclouds2001/mdn-tools/blob/master/results/incorrect_mdn_url.json

per spec https://drafts.csswg.org/css-ui/#propdef-input-security, it seems that it'll be removed from spec soon, see also https://github.com/w3c/csswg-drafts/issues/6788 and https://github.com/whatwg/html/issues/7293; also notice that per https://github.com/mdn/browser-compat-data/pull/12822, it only support in safari preview and disabled by default via a runtime flag

Relates to https://github.com/mdn/browser-compat-data/issues/12664
Relates to https://github.com/mdn/data/issues/511
Relates to https://github.com/mdn/content/issues/9422
Relates to https://github.com/mdn/content/issues/10438


bsmth:
Thanks a lot, I found the discussions about this really hard to follow, and I had to revisit it a few times to understand the context, but the important piece, like you mentioned, is the spec section:

https://drafts.csswg.org/css-ui/#input-security

So I'm okay with removing this

css/properties.json
```diff
@@ -6119,21 +6119,6 @@
...
-    "syntax": "auto | none",
-    "media": "interactive",
-    "inherited": false,
-    "animationType": "byComputedValueType",
-    "percentages": "no",
-    "groups": [
-      "CSS Basic User Interface"
-    ],
-    "initial": "auto",
-    "appliesto": "sensitiveTextInputs",
```
bsmth, line 6132: Should we also be removing this

```
./css/properties.schema.json:        "sensitiveTextInputs",
```

Seems like it's not used anywhere else

skyclouds2001: removed!

bsmth: ty

bsmth (approved):

PR #864, opened 2024-12-29
feat(css): add legacy `-webkit-user-select` property

skyclouds2001:
### Description

the `-webkit-user-select` property is a vendor-prefixed alias for `user-select` property, note that per bcd, the vendor-prefixed alias is supported in all major browsers, and Safari only support the vendor-prefixed alias, so it is worth to add it back

also, remove the not supported `contain` value from `user-select` property syntax

reported in https://github.com/frenic/csstype/issues/191

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/user-select
https://drafts.csswg.org/css-ui/#propdef-user-select

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -1686,6 +1686,22 @@
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
+  "-webkit-user-select": {
+    "syntax": "auto | text | none | all",
+    "media": "visual",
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Thank you! It would be nice to be able to alias / mirror values instead of duplicating, but that's for another day 👍🏻

PR #865, opened 2024-12-29
feat(css): consistent naming for function-like pseudo-classes

### LINKED ISSUE #242: Inconsistent naming for pseudo-classes that act as functions
I'm not sure if there's a better name for "pseudo-classes that act as functions"?

Essentially what I mean are pseudo-classes like `:dir()`, `:lang()`, etc.

This is a list of function selectors with weird/incorrect names in [`selectors.json`](https://github.com/mdn/data/blob/f9a599abe52505a26dfd5215b84e0df46e597a2a/css/selectors.json):

- [`:any`](https://github.com/mdn/data/blob/f9a599abe52505a26dfd5215b84e0df46e597a2a/css/selectors.json#L82): This is weird because it only has syntax specifically for the `-webkit-`/`-moz-` vendor prefixed versions. Also, it should be a function like `:any()`.
- [`:dir`](https://github.com/mdn/data/blob/f9a599abe52505a26dfd5215b84e0df46e597a2a/css/selectors.json#L122): Should be `:dir()`.
- [`:lang`](https://github.com/mdn/data/blob/f9a599abe52505a26dfd5215b84e0df46e597a2a/css/selectors.json#L274): Should be `:lang()`.
- [`:not`](https://github.com/mdn/data/blob/f9a599abe52505a26dfd5215b84e0df46e597a2a/css/selectors.json#L315): Should be `:not()`.
- [`:nth-child`](https://github.com/mdn/data/blob/f9a599abe52505a26dfd5215b84e0df46e597a2a/css/selectors.json#L323): Should be `:nth-child()`.
- [`:nth-last-child`](https://github.com/mdn/data/blob/f9a599abe52505a26dfd5215b84e0df46e597a2a/css/selectors.json#L331): Should be `:nth-last-child()`.
- [`:nth-last-of-type`](https://github.com/mdn/data/blob/f9a599abe52505a26dfd5215b84e0df46e597a2a/css/selectors.json#L339): Should be `:nth-last-of-type()`.
- [`:nth-of-type`](https://github.com/mdn/data/blob/f9a599abe52505a26dfd5215b84e0df46e597a2a/css/selectors.json#L347): Should be `:nth-of-type()`.

Oddly, `:host()` and `:host-context()` are fine in this regard.

The logic that any selectors with `()` should have `()` included in their name is that there are selectors like `:host` as well as `:host()`, so you can't just take off the `()` or those two would conflict.

skyclouds2001:
### Description

consistent naming for function-like pseudo-classes

Fixes https://github.com/mdn/data/issues/242


bsmth (approved):
Thanks a lot. These all make sense to me. I wonder if we need to update the titles in the `::view-transition-*` docs in mdn/content because I agree the syntax is function-like. Also:

https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API#pseudo-elements

skyclouds2001:
> These all make sense to me. I wonder if we need to update the titles in the `::view-transition-*` docs in mdn/content because I agree the syntax is function-like.

yes, I agree to do the same thing in mdn/content

PR #677, opened 2023-07-12
Remove CSS at-rule @viewport

mems:
### Description

Remove CSS at-rule @viewport

### Motivation

This at-rule has been removed from spec CSS Viewport Module Level 1

### Additional details

https://github.com/w3c/csswg-drafts/pull/7670

### Related issues and pull requests

N/A

PR #870, opened 2025-01-01
chore(css): remove not-supported values for `text-spacing-trim`

skyclouds2001:
### Description

the `trim-both` `trim-all` `auto` values are in spec but not supported by any platform yet

see spec https://drafts.csswg.org/css-text-4/#propdef-text-spacing-trim and mdn docs https://developer.mozilla.org/en-US/docs/Web/CSS/text-spacing-trim

### Additional details

see also https://github.com/skyclouds2001/mdn-tools/blob/master/docs/not-fully-supported-feature.md

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -10105,7 +10105,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-size-adjust"
   },
   "text-spacing-trim": {
-    "syntax": "space-all | normal | space-first | trim-start | trim-both | trim-all | auto",
+    "syntax": "space-all | normal | space-first | trim-start",
     "media": "visual",
     "inherited": true,
     "animationType": "discrete",


bsmth:
Thank you! While this is correct, I'm curious if consumers want all spec values even if not supported yet. I can imagine cases where a user wants to know possible property values according to spec, even if implementations haven't caught up. Additionally, we won't have to revisit when these are actually added (in BCD, for instance), or we won't forget to add at a later point.

So I see two cases where mdn/data and bcd diverge:

- removed (something is being removed from the web platform)
- not yet supported

I think for the not-yet supported things, we may leave the entries in here. Something is being removed, I think it's safe to delete.

skyclouds2001:
> While this is correct, I'm curious if consumers want all spec values even if not supported yet. I can imagine cases where a user wants to know possible property values according to spec, even if implementations haven't caught up.

I think in those cases, consumers is better to use [w3c/webref](https://github.com/w3c/webref) for the latest data directly offered from the spec. See also https://github.com/mdn/data/issues/585.


Other refereneces:

- https://github.com/orgs/mdn/discussions/716

skyclouds2001:
Also notice that such cases existed for removing not-supported values of CSS features, like https://github.com/mdn/data/pull/775, https://github.com/mdn/data/pull/677, https://github.com/mdn/data/pull/417, https://github.com/mdn/data/pull/32 and more.

bsmth:
> I think it is better to follow the similar behavior like mdn/browser-compat-data - only keep values that has been supported by at least one engine.

Thanks a lot for the thoughtful response, I think the points you're making are convincing. For now, I'd like to merge to unblock this cleanup work, but it would be great to get some guidelines in place (in this repo) for cases like this. The opinions of consumers would mostly guide this decision, so I hope we get some input on that soon.

bsmth (approved):

Issue #230, opened 2018-05-14
Simplify syntaxes


lahmatiy:
While working on new matching approach in [CSSTree](https://github.com/csstree/csstree) I found that some syntaxes can be simplified. I wrote [a script](https://github.com/csstree/csstree/blob/new-match/scripts/can-be-simplified.js) to detect some cases. Here is the output:

```
# mdn/atRules/@font-face "font-display"
A top level group can be omitted
Current syntax:   [ auto | block | swap | fallback | optional ]
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Suggested syntax: auto | block | swap | fallback | optional
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

# mdn/atRules "@import"
A single term group can be omitted and a multiplier moved to a term
Current syntax:   @import [ <string> | <url> ] [ <media-query-list> ]? ;
                                               ~~~~~~~~~~~~~~~~~~~~~~~
Suggested syntax: @import [ <string> | <url> ] <media-query-list>? ;
                                               ~~~~~~~~~~~~~~~~~~~

# mdn/properties "cursor"
A top level group can be omitted
Current syntax:   [ [ <url> [ <x> <y> ]? , ]* [ auto | default | none | context-menu | help | pointer | progress | wait | cell | crosshair | text | vertical-text | alias | copy | move | no-drop | not-allowed | e-resize | n-resize | ne-resize | nw-resize | s-resize | se-resize | sw-resize | w-resize | ew-resize | ns-resize | nesw-resize | nwse-resize | col-resize | row-resize | all-scroll | zoom-in | zoom-out | grab | grabbing ] ]
[...]

PR #877, opened 2025-01-03
feat(css): Update syntax for `outline`

skyclouds2001:
### Description

see https://drafts.csswg.org/css-ui/#propdef-outline
reported in https://github.com/mdn/data/issues/230


pepelsbey (approved):
Looks good! Thank you :)

PR #880, opened 2025-01-04
chore(css): remove never used CSS Break group

skyclouds2001:
### Description

this group is never used in repo, so remove


bsmth (approved):
Thanks!

PR #881, opened 2025-01-04
chore(css): Update group of CSS Cascading and Inheritance

skyclouds2001:
### Description

the `all` property and the `@import` at-rule should also belongs to CSS Cascading and Inheritance

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade
https://drafts.csswg.org/css-cascade/
https://drafts.csswg.org/css-cascade-5/


pepelsbey (commented):

PR #882, opened 2025-01-05
chore(css): Remove CSS Miscellaneous group & Rename CSS Variables group

skyclouds2001:
### Description]

remove *CSS Miscellaneous* group as it is not the real group, and set those features to its real values

https://developer.mozilla.org/en-US/docs/Web/CSS/text-rendering
https://svgwg.org/svg2-draft/painting.html#TextRenderingProperty

https://developer.mozilla.org/en-US/docs/Web/CSS/all
https://drafts.csswg.org/css-cascade/#propdef-all

https://developer.mozilla.org/en-US/docs/Web/CSS/var
https://drafts.csswg.org/css-variables/#funcdef-var

https://developer.mozilla.org/en-US/docs/Web/CSS/element
https://drafts.csswg.org/css-images-4/#funcdef-element

https://drafts.csswg.org/css-content/#target-text-function
https://drafts.csswg.org/css-content/#funcdef-content-leader


bsmth:
I guess "Variables" is "CSS Custom Properties for Cascading Variables" per the module name, but I'm not sure how strict this needs to be.

bsmth (approved):
Everything looks good except for Variables comment, but +1 either way

skyclouds2001:
emm, seems mdn web docs also use "CSS Custom Properties for Cascading Variables"

see https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascading_variables

PR #884, opened 2025-01-05
chore(css): Remove never used CSS Regions group

skyclouds2001:
### Description

the group is never used, so remove it


bsmth (approved):
Thank you!

PR #885, opened 2025-01-05
feat(css): Update syntax for `path()` & `polygon()`

skyclouds2001:
### Description

see https://drafts.csswg.org/css-shapes/#funcdef-basic-shape-polygon and https://drafts.csswg.org/css-shapes/#funcdef-basic-shape-path

partially reported in https://github.com/mdn/data/issues/230


pepelsbey (approved):
Looks good! Thank you :)

PR #886, opened 2025-01-05
feat(css): Update syntax of filter functions for optional parameters

skyclouds2001:
### Description

same as title, also limit those functions' group to *Filter Effects* only

spec at https://drafts.fxtf.org/filter-effects/#typedef-filter-function

reported in https://github.com/mdn/data/issues/230


bsmth (approved):
great, thank you!

PR #887, opened 2025-01-06
feat(css): Add `@view-transition/types`

skyclouds2001:
### Description

https://drafts.csswg.org/css-view-transitions-2/#descdef-view-transition-types

test in chrome shows it is supported, though not list in bcd or mdn docs


pepelsbey (approved):
Looks good! Thank you :)

Issue #218, opened 2018-04-30
CSS seems to be missing "speak"


tommai78101:
Here's where the CSS standard is defined:

https://www.w3.org/TR/CSS21/aural.html#speaking-props

It's missing the following:

```
.foo{
  speak: none;
  speak: normal;
  speak: spell-out
}
```

frenic:
This seems to be the latest revisions:
- https://www.w3.org/TR/css3-speech/#speaking-props-speak
- https://drafts.csswg.org/css-speech-1/#speak.

`pause`, `cue`, `rest` etc. are missing too. Not sure which browsers supports these properties though.

tommai78101:
Yeah, even caniuse.com couldn't help on this.

frenic:
Edge and Firefox doesn't seem to know `speak` or any other of the CSS Speech properties. Chrome however knows `speak`. But some values refers to the [`speak-as` spec](https://drafts.csswg.org/css-speech-1/#speak-as), not the official `speak`. I've no clue why, if this comes from an early spec or if it's a bug.

![image](https://user-images.githubusercontent.com/237182/39451974-dfa0a5dc-4cd0-11e8-94b2-2dadde050126.png)

tommai78101:
If it's from an earlier spec, we can go with the latest revision, and keep it at `auto`, `never`, and `always` for now.

Since Chrome has it, maybe Firefox and Edge should have it?

cmxl:
Adoption seems to be pretty wide-spread:
https://caniuse.com/?search=speak

Some update on this issue?

SebastianZ:
I've dug a little into this topic.

## Specifications

There are actually two different definitions for `speak`. [In CSS 2.1](https://www.w3.org/TR/CSS21/aural.html#speaking-props) (and [CSS 2.2](https://www.w3.org/TR/CSS22/aural.html#speaking-props)) it is defined as
```
normal | none | spell-out | inherit
```

And in [CSS Speech 1](https://www.w3.org/TR/css-speech-1/#background) it is
```
auto | never | always
```

CSS Speech 1 additionally defines a [`speak-as` property](https://www.w3.org/TR/css-speech-1/#speaking-props-speak-as) which has the following values
```
normal | spell-out || digits || [ literal-punctuation | no-punctuation ]
```

Additionally, CSS 2 has this note at the top of that section:

> This chapter is informative. UAs are not required to implement the properties of this chapter in order to conform to CSS 2.2.

[...]

PR #889, opened 2025-01-07
feat(css): Add data for experimental `speak-as` property

skyclouds2001:
### Description

note the `speak-as` property is not the same as the `@counter-style/speak-as` descriptor

test for that property is in https://github.com/mdn/data/issues/218#issuecomment-2571535303

see https://github.com/mdn/browser-compat-data/blob/main/css/properties/speak-as.json for bcd, and only Safari support this feature t present

also refers to https://drafts.csswg.org/css-speech/#propdef-speak-as for spec definition


pepelsbey (approved):
Looks good! Thank you :)

PR #890, opened 2025-01-07
chore(css): Update for CSS Generated Content group

skyclouds2001:
### Description

set group to "CSS Generated Content" for features of CSS Generated Content group
set group to correct value for features not of CSS Generated Content group

### Additional details

https://drafts.csswg.org/css-content/ for spec
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_generated_content for mdn web docs


pepelsbey (approved):
Looks good! Thank you :)

css/definitions.json
```diff
@@ -56,6 +56,8 @@
       "CSS Transitions",
       "CSS Types",
       "CSS Units",
+      "CSS Values and Units",
+      "CSS Variables",
```
skyclouds2001, line 60:
```suggestion
```

PR #891, opened 2025-01-07
chore(css): Update group for `paint()` function

skyclouds2001:
### Description

https://drafts.css-houdini.org/css-paint-api/
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Houdini


pepelsbey (approved):
Looks good! Thank you :)

PR #892, opened 2025-01-08
feat(css): Add CSS Shadow Parts group to `::part()`

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_shadow_parts
https://developer.mozilla.org/en-US/docs/Web/CSS/::part
https://drafts.csswg.org/css-shadow-parts/#selectordef-part


pepelsbey (approved):
Looks good! Thank you :)

PR #897, opened 2025-01-13
feat(css): Update `filter-function-list` to `filter-value-list`

skyclouds2001:
### Description

the spec has renamed this syntax

https://drafts.fxtf.org/filter-effects/#propdef-filter
https://drafts.fxtf.org/filter-effects-2/#propdef-backdrop-filter

https://github.com/mdn/content/pull/37633 for content update


pepelsbey (approved):
Looks good! Thank you :)

PR #898, opened 2025-01-17
feat(css): Update group value for CSS Lists and Counters group

skyclouds2001:
### Description

these features are of CSS Lists and Counters group

### Additional details

https://drafts.csswg.org/css-lists/ for spec
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_lists for mdn web docs


pepelsbey (approved):
Looks good! Thank you :)

PR #899, opened 2025-01-18
style(css): fix feature order

skyclouds2001:
### Description

reported in https://github.com/skyclouds2001/mdn-tools/blob/master/results/unordered_data.json


pepelsbey (approved):
Looks good! Thank you :)

PR #900, opened 2025-01-18
chore(css): Rename CSS Device Adaptation to CSS Viewport

skyclouds2001:
### Description

see https://drafts.csswg.org/css-viewport/ for spec


pepelsbey (approved):
Looks good! Thank you :)

PR #901, opened 2025-01-18
feat(css): Update synatx for css `zoom` property

skyclouds2001:
### Description

see comments below

see https://drafts.csswg.org/css-viewport/#propdef-zoom for spec

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -10853,18 +10853,18 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/z-index"
   },
   "zoom": {
-    "syntax": "normal | reset | <number> | <percentage>",
+    "syntax": "normal | reset | <number [0,∞]> || <percentage [0,∞]>",
     "media": "visual",
     "inherited": false,
     "animationType": "notAnimatable",
-    "percentages": "no",
+    "percentages": "convertedToNumber",
     "groups": [
       "Microsoft Extensions"
     ],
-    "initial": "normal",
+    "initial": "1",
     "appliesto": "allElements",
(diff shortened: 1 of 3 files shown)


css/properties.json
```diff
@@ -10832,18 +10832,18 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/z-index"
   },
   "zoom": {
-    "syntax": "normal | reset | <number> | <percentage>",
+    "syntax": "normal | reset | <number [0,∞]> || <percentage [0,∞]>",
```
skyclouds2001, line 10835: the syntax for the spec is `<number [0,∞]> || <percentage [0,∞]>`, while the `normal` keyword is supported in chrome per manually test; `reset` keyword is supported in safari per bcd and manually test

notice since *double bar has precedence over the single bar*, *normal | reset | <number [0,∞]> || <percentage [0,∞]>* is equal to *normal | reset | [ <number [0,∞]> || <percentage [0,∞]> ]*

css/properties.json
```diff
@@ -10832,18 +10832,18 @@
...
     "inherited": false,
     "animationType": "notAnimatable",
-    "percentages": "no",
+    "percentages": "convertedToNumber",
     "groups": [
       "Microsoft Extensions"
     ],
-    "initial": "normal",
+    "initial": "1",
     "appliesto": "allElements",
```
skyclouds2001, line 10844: while spec says it as *all <length> property values of all elements*, this should just be *all elements*

pepelsbey (approved):
Looks good! Thank you :)

PR #902, opened 2025-01-19
chore(css): Rename CSS Motion Path to Motion Path

skyclouds2001:
### Description

thi match the behavior of other filter modules

also remove *Motion Path* from path()'s group


pepelsbey (approved):
Looks good! Thank you :)

PR #903, opened 2025-01-19
chore(css): Remove cross-fade from Filter Effects group

skyclouds2001:
### Description

it is not part of the spec at all

see spec https://drafts.fxtf.org/filter-effects/ and css module docs https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_filter_effects


pepelsbey (commented):

PR #904, opened 2025-01-19
chore(css): Remove never used CSSOM View group

skyclouds2001:
### Description

this item is never used so can be removed


pepelsbey (approved):
Looks good! Thank you :)

PR #905, opened 2025-01-19
feat(css): Add css `symbols()` function

skyclouds2001:
### Description

see https://developer.mozilla.org/en-US/docs/Web/CSS/symbols and https://drafts.csswg.org/css-counter-styles/#funcdef-symbols


pepelsbey (approved):
Looks good! Thank you :)

PR #907, opened 2025-01-22
feat(css): Update syntax for `an+b` type

skyclouds2001:
### Description

see https://drafts.csswg.org/selectors/#nth-last-of-type-pseudo and https://drafts.csswg.org/css-syntax-3/#anb-microsyntax

such syntax has been used by selectors like `:nth-child()` `:nth-last-child()` `:nth-last-of-type()` `:nth-of-type()`


pepelsbey (approved):
Looks good! Thank you :)

PR #909, opened 2025-01-25
feat(css): Update status for page-break-* properties

skyclouds2001:
### Description

see also bcd change at https://github.com/mdn/browser-compat-data/pull/25745


pepelsbey (approved):
Looks good! Thank you :)

PR #910, opened 2025-01-25
chore(css): Fix group key for `text-overflow`

skyclouds2001:
### Description

it should be CSS Overflow, not CSS Basic User Interface


pepelsbey (approved):
Looks good! Thank you :)

PR #912, opened 2025-01-28
chore(css): Add align-* and justify-* properties with CSS Flexible Bo…

skyclouds2001:
…x Layout group

### Description

see https://drafts.csswg.org/css-flexbox/ and https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout


bsmth (approved):
LGTM, thank you!

PR #914, opened 2025-01-28
chore(css): FIx group for CSS Box Sizing

skyclouds2001:
### Description

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_box_sizing
https://drafts.csswg.org/css-sizing-3/
https://drafts.csswg.org/css-sizing-4/

### DIFF
--- a/css/functions.json
+++ b/css/functions.json
@@ -230,7 +230,7 @@
   "fit-content()": {
     "syntax": "fit-content( <length-percentage [0,∞]> )",
     "groups": [
-      "CSS Box Sizing"
+      "CSS Grid Layout"
     ],
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/fit-content_function"
--- a/css/functions.json
+++ b/css/functions.json
@@ -230,6 +230,7 @@
   "fit-content()": {
     "syntax": "fit-content( <length-percentage [0,∞]> )",
     "groups": [
+      "CSS Box Sizing",
       "CSS Grid Layout"
     ],
     "status": "standard",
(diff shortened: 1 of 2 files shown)


css/functions.json
```diff
@@ -230,7 +230,7 @@
   "fit-content()": {
     "syntax": "fit-content( <length-percentage [0,∞]> )",
     "groups": [
-      "CSS Box Model"
+      "CSS Box Sizing"
```
bsmth, line 233: Isn't this Grid Layout?

bsmth:
Everything looks good, one comment about `fit-content()`

css/functions.json
skyclouds2001, line 233:
```suggestion
      "CSS Grid Layout"
```

skyclouds2001: yes! my mistake

css/functions.json
```diff
@@ -230,7 +230,7 @@
   "fit-content()": {
     "syntax": "fit-content( <length-percentage [0,∞]> )",
     "groups": [
-      "CSS Box Model"
+      "CSS Grid Layout"
```
skyclouds2001, line 233:
```suggestion
      "CSS Box Sizing",
      "CSS Grid Layout"
```

css/functions.json
skyclouds2001: it is both defined in CSS Box Sizing module and CSS Grid Layout module, so both need to be added

see https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content_function, https://drafts.csswg.org/css-sizing/#funcdef-width-fit-content and https://drafts.csswg.org/css-grid/#funcdef-grid-template-columns-fit-content

bsmth (approved):
Thank you!

PR #915, opened 2025-01-28
chore(css): Fix group for css `env()` function

skyclouds2001:
### Description

see https://developer.mozilla.org/en-US/docs/Web/CSS/env and https://drafts.csswg.org/css-env/, this should be CSS Environment Variables Module


pepelsbey (approved):
Looks good! Thank you :)

PR #916, opened 2025-01-28
chore(css): Fix group for CSS Display module features

skyclouds2001:
### Description

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_display
https://drafts.csswg.org/css-display-4/


pepelsbey (approved):
Looks good! Thank you :)

PR #917, opened 2025-01-30
chore(css): Fix group value for CSS Overscroll Behavior features

skyclouds2001:
### Description

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_overscroll_behavior
https://drafts.csswg.org/css-overscroll/


bsmth (approved):
Looks great, tnx 👍🏻

PR #918, opened 2025-01-30
chore(css): Rename CSS Pages group to CSS Paged Media group

skyclouds2001:
### Description

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_paged_media
https://drafts.csswg.org/css-page/


bsmth (approved):
LGTM thanks!

PR #919, opened 2025-01-30
chore(css): Rename CSS Columns to CSS Multi-column Layout

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_multicol_layout
https://drafts.csswg.org/css-multicol/


bsmth (approved):
great, thank you!

PR #921, opened 2025-02-06
chore(css): Fix group for some CSS Inline module feature

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_inline_layout
https://drafts.csswg.org/css-inline/


pepelsbey (approved):
Looks good! Thank you :)

PR #922, opened 2025-02-07
chore(css): Add CSS Anchor Positioning group

skyclouds2001:
### Description

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning
https://drafts.csswg.org/css-anchor-position/


pepelsbey (approved):
Looks good! Thank you :)

PR #925, opened 2025-02-09
chore(css): Fix group for `clip-rule`

skyclouds2001:
### Motivation

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_masking


pepelsbey (approved):
Looks good! Thank you :)

PR #926, opened 2025-02-09
chore(css): Fix group for `shape-rendering`

skyclouds2001:
### Motivation

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_shapes
https://developer.mozilla.org/en-US/docs/Web/CSS/shape-rendering


pepelsbey (approved):
Looks good! Thank you :)

PR #927, opened 2025-02-09
feat(css): Add missing `rect()` and `xywh()` functions

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/rect
https://drafts.csswg.org/css-shapes-1/#funcdef-basic-shape-rect
https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/xywh
https://drafts.csswg.org/css-shapes-1/#funcdef-basic-shape-xywh

### DIFF
--- a/css/functions.json
+++ b/css/functions.json
@@ -517,6 +517,14 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/ray"
   },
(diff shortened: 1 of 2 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #929, opened 2025-02-09
chore(css): Rename CSS Logical Properties to CSS Logical Properties and Values

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values
https://drafts.csswg.org/css-logical/


pepelsbey (approved):
Looks good! Thank you :)

PR #930, opened 2025-02-11
chore(css): Fix group for CSS Transitions module

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_transitions
https://drafts.csswg.org/css-transitions/
https://drafts.csswg.org/css-transitions-2/


pepelsbey (commented):

PR #931, opened 2025-02-11
chore(css): Update group for CSS Animations

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animations
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scroll-driven_animations
https://drafts.csswg.org/css-animations/
https://drafts.csswg.org/css-animations-2/
https://drafts.csswg.org/scroll-animations/

### DIFF
--- a/css/at-rules.json
+++ b/css/at-rules.json
@@ -521,7 +521,7 @@
       "CSSStartingStyleRule"
     ],
     "groups": [
-      "CSS Animations"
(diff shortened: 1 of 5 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #932, opened 2025-02-17
chore(css): Update group for CSS Fonts

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts
https://drafts.csswg.org/css-fonts/
https://drafts.csswg.org/css-fonts-5/


pepelsbey (approved):
Looks good! Thank you :)

PR #933, opened 2025-02-17
chore(css): Update group for CSS Text

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_text
https://drafts.csswg.org/css-text/
https://drafts.csswg.org/css-text-4/
https://drafts.csswg.org/css-size-adjust/


pepelsbey (approved):
Looks good! Thank you :)

PR #936, opened 2025-02-25
chore(css): Update status for `:has-slotted`

skyclouds2001:
### Description

see https://github.com/mdn/browser-compat-data/blob/main/css/selectors/has-slotted.json and https://github.com/skyclouds2001/mdn-tools/blob/master/data/results/mismatch_status.json


bsmth (approved):
Nice, tnx 👍🏻

PR #937, opened 2025-02-25
chore(css): Add mdn_url for newly documented features

skyclouds2001:
### Description

see https://github.com/skyclouds2001/mdn-tools/blob/master/data/results/incorrect_mdn_url.json


bsmth (approved):
TY!

PR #938, opened 2025-02-25
feat(css): Sync syntax for some css functions

skyclouds2001:
### Description

see https://github.com/skyclouds2001/mdn-tools/blob/master/data/results/inconsistent_function.json


bsmth (approved):
Looks good, tnx

PR #940, opened 2025-03-08
feat(css): Add data for `view-transition-class`

skyclouds2001:
### Description

the feature is supported in chrome and safari

### Additional details

https://github.com/mdn/browser-compat-data/blob/main/css/properties/view-transition-class.json
https://drafts.csswg.org/css-view-transitions-2/#view-transition-class-prop


bsmth (approved):
Nice, LGTM, thanks 👍🏻

PR #943, opened 2025-03-19
chore(css): Sync features's status with bcd

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/::details-content
https://developer.mozilla.org/en-US/docs/Web/CSS/font-stretch
https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-stretch


pepelsbey (approved):
Looks good! Thank you :)

PR #945, opened 2025-03-22
feat(css): Update syntax for `font-size`

skyclouds2001:
### Description

add support for the new keyword `math`, which is supported in chrome and firefox for more than 2 years, also adjust some information field to match the spec

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/font-size
https://drafts.csswg.org/css-fonts/#propdef-font-size

Relates to https://github.com/stylelint/stylelint/issues/8436 and https://github.com/stylelint/stylelint/pull/8489

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5108,18 +5108,18 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/font-palette"
   },
   "font-size": {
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #946, opened 2025-03-22
feat(css): Update syntax for `view-timeline`

skyclouds2001:
### Motivation

Fix syntax for `view-timeline` to add support for `view-timeline-inset` as a value, and fix the syntax for `view-timeline-name` as `none` keyword vaue cna also accept multiple values

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/view-timeline
https://drafts.csswg.org/scroll-animations/#propdef-view-timeline
https://developer.mozilla.org/en-US/docs/Web/CSS/view-timeline-name
https://drafts.csswg.org/scroll-animations/#propdef-view-timeline-name

Relates to https://github.com/stylelint/stylelint/issues/8100

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -10553,7 +10553,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/vertical-align"
   },
   "view-timeline": {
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #947, opened 2025-03-23
feat(css): Update syntax for `*-conic-gradient()`

skyclouds2001:
### Description

using `<conic-gradient-syntax>` to both present `conic-gradient()` and `repeat-conic-gradient()`
add `<color-interpolation-method>` to support for the new interpolation color space syntax, which is supported in all major browsers
also add `<zero>` to present `0` generic value

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/conic-gradient
https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/repeat-conic-gradient
https://drafts.csswg.org/css-images-4/#conic-gradients
https://drafts.csswg.org/css-images-4/#repeating-gradients

Relates to https://github.com/stylelint/stylelint/issues/8347

### DIFF
--- a/css/functions.json
+++ b/css/functions.json
@@ -136,7 +136,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/color-mix"
   },
   "conic-gradient()": {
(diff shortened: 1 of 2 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #948, opened 2025-03-23
feat(css): Update syntax for `*-linear-gradient()`

skyclouds2001:
### Description

using `<linear-gradient-syntax>` to both present `linear-gradient()` and `repeat-linear-gradient()`
add `<color-interpolation-method>` to support for the new interpolation color space syntax, which is supported in all major browsers
also add `<zero>` to present `0` generic value

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient
https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/repeating-linear-gradient
https://drafts.csswg.org/css-images-4/#linear-gradients
https://drafts.csswg.org/css-images-4/#repeating-gradients

Relates to https://github.com/stylelint/stylelint/issues/8347

### DIFF
--- a/css/functions.json
+++ b/css/functions.json
@@ -355,7 +355,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/light-dark"
   },
   "linear-gradient()": {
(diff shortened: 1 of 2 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #951, opened 2025-03-23
feat(css): Update syntax for `<easing-function>`

skyclouds2001:
### Description

add *CSS Easing Functions* group and assgin `<easing-function>` to that group
extract `cubic-bezier()` function from `<cubic-bezier-timing-function>`, and update syntax of `cubic-bezier()` to match the spec, which is a simplify
renaming `<step-timing-function>` to be align with the spec, and extract `steps()` from `<step-timing-function>`
update syntax of `<easing-function>` to better represent `linear()` function, which also extract `<linear-easing-function>`

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function
https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function/cubic-bezier
https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function/linear
https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function/steps

https://drafts.csswg.org/css-easing-2/#typedef-easing-function
https://drafts.csswg.org/css-easing-2/#typedef-linear-easing-function
https://drafts.csswg.org/css-easing-2/#typedef-cubic-bezier-easing-function
https://drafts.csswg.org/css-easing-2/#typedef-step-easing-function

Relates to https://github.com/stylelint/stylelint/issues/8379

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -300,7 +300,7 @@
     "syntax": "drop-shadow( [ <color>? && <length>{2,3} ] )"
   },
   "easing-function": {
-    "syntax": "<linear-easing-function> | <cubic-bezier-timing-function> | <step-easing-function>"
+    "syntax": "<linear-easing-function> | <cubic-bezier-easing-function> | <step-easing-function>"
   },
   "east-asian-variant-values": {
     "syntax": "[ jis78 | jis83 | jis90 | jis04 | simplified | traditional ]"
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -254,8 +254,11 @@
   "cross-fade()": {
     "syntax": "cross-fade( <cf-mixing-image> , <cf-final-image>? )"
   },
+  "cubic-bezier()": {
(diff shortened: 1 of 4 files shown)


css/syntaxes.json
```diff
@@ -297,7 +300,7 @@
     "syntax": "drop-shadow( [ <color>? && <length>{2,3} ] )"
   },
   "easing-function": {
-    "syntax": "linear | <cubic-bezier-timing-function> | <step-timing-function>"
+    "syntax": "<linear-easing-function> | <cubic-bezier-timing-function> | <step-easing-function>"
```
dipikabh, line 303:
```suggestion
    "syntax": "<linear-easing-function> | <cubic-bezier-easing-function> | <step-easing-function>"
```

dipikabh (changes requested):
Thank you so much for adding these! 🎉 

just a very tiny nit that could be fixed

dipikabh (approved):
Thanks a lot!!

PR #952, opened 2025-03-26
feat(css): Update syntax for `<generic-family>`

skyclouds2001:
### Description

the `<generic-family>` is used by `font-family`, update it to match the latest spec and browser implement:

split `<generic-family>` into `<generic-complete>` and `<generic-incomplete>`, add support for non-standatd `emoji` and `fangsong` value

also remove the `<generic-name>` type which is never used

### Motivation

https://drafts.csswg.org/css-fonts/#propdef-font-family

### DIFF
--- a/css/syntaxes.json
+++ b/css/syntaxes.json
@@ -380,14 +380,17 @@
   "frequency-percentage": {
     "syntax": "<frequency> | <percentage>"
   },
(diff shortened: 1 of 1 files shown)


dipikabh (approved):
Looks good, thank you!

PR #954, opened 2025-03-26
feat(css): Update syntax for `@keyframes`

skyclouds2001:
### Description

change `@keyframes`'s at-rule body from `<keyframe-block-list>` to `<qualified-rule-list>`
update `<keyframe-selector>`'s range limit per spec

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes
https://drafts.csswg.org/css-animations/#at-ruledef-keyframes

Relates to https://github.com/stylelint/stylelint/issues/7855

### DIFF
--- a/css/at-rules.json
+++ b/css/at-rules.json
@@ -340,7 +340,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@import"
   },
   "@keyframes": {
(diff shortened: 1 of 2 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #955, opened 2025-03-26
feat(css): Update syntax for `container-type`

skyclouds2001:
### Description

chrome 133 add support for `scroll-state` keyword

so it's time to update it to match the spec

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/container-type
https://www.w3.org/TR/css-conditional-5/#propdef-container-type

Relates to https://github.com/csstree/csstree/pull/313


pepelsbey (approved):
Looks good! Thank you :)

PR #956, opened 2025-03-28
feat(css): Add data for `alignment-baseline`

skyclouds2001:
### Description

https://developer.mozilla.org/en-US/docs/Web/CSS/alignment-baseline
https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/alignment-baseline
https://www.w3.org/TR/css-inline-3/#propdef-alignment-baseline
https://svgwg.org/svg2-draft/text.html#AlignmentBaselineProperty

spec defines `baseline | text-bottom | alphabetic | ideographic | middle | central | mathematical | text-top`, while `text-bottom` and `text-top` keywords are not supportes per BCD; also add deprecated & non-standard `auto` `before-edge`  `after-edge` `text-before-edge` `text-after-edge` `hanging` keywords


css/properties.json
```diff
@@ -1720,6 +1720,22 @@
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
+  "alignment-baseline": {
+    "syntax": "baseline | text-bottom | alphabetic | ideographic | middle | central | mathematical | text-top | auto | before-edge | after-edge | text-before-edge | text-after-edge | hanging",
```
skyclouds2001, line 1724:
```suggestion
    "syntax": "baseline | alphabetic | ideographic | middle | central | mathematical | auto | before-edge | after-edge | text-before-edge | text-after-edge | hanging",
```

css/properties.json
```diff
@@ -1720,6 +1720,22 @@
     "order": "uniqueOrder",
     "status": "nonstandard"
   },
+  "alignment-baseline": {
+    "syntax": "baseline | alphabetic | ideographic | middle | central | mathematical | auto | before-edge | after-edge | text-before-edge | text-after-edge | hanging",
```
dipikabh, line 1724: I think the syntax should only include values in the current spec. We shouldn't list deprecated keywords. The good thing is that the content page already has a note to capture the deprecated list.
As far as BCD is concerned, we can determine if an update is needed.

css/properties.json
```diff
@@ -1720,6 +1720,22 @@
...
+    "media": "none",
+    "inherited": false,
+    "animationType": "discrete",
+    "percentages": "no",
+    "groups": [
+      "CSS Inline"
+    ],
+    "initial": "baseline",
+    "appliesto": "inlineLevelBoxesFlexItemsGridItemsTableCellsAndSVGTextContentElements",
```
dipikabh, line 1733: I don't see this value defined in css/properties.schema.json. Could you mention the source of your info?

css/properties.json
```diff
@@ -1720,6 +1720,22 @@
...
+    "computed": "specifiedKeyword",
```
dipikabh, line 1734:
```suggestion
    "computed": "theSpecifiedKeyword",
```

css/properties.schema.json
```diff
@@ -169,6 +169,7 @@
         "sameAsMinWidthAndMinHeight",
         "sameAsWidthAndHeight",
         "specifiedIntegerOrAbsoluteLength",
+        "specifiedKeyword",
```
dipikabh, line 172: this is perhaps not needed since we have:

https://github.com/mdn/data/blob/2e6ee0843a895ca7d71485b4899a53f8746ec430/css/properties.schema.json#L177

css/properties.schema.json
```diff
@@ -248,9 +249,10 @@
         "iframeElements",
         "images",
         "inFlowBlockLevelElements",
-        "inFlowChildrenOfBoxElements",
         "inlineBoxesAndBlockContainers",
+        "inFlowChildrenOfBoxElements",
```
dipikabh, line 253: Sorry, I can't make out why this change was needed

l10n/css.json
```diff
@@ -1725,6 +1728,9 @@
     "ja": "指定された整数値または絶対的な長さ",
     "ru": "указанное целое число или абсолютная длина"
   },
+  "specifiedKeyword": {
+    "en-US": "specified keyword"
+  },
```
dipikabh, line 1733: might not be needed

css/properties.json
dipikabh: Okay, I see that you're adding an entry for this in the schema file

dipikabh (changes requested):
Thanks @skyclouds2001. I've added a few comments for your consideration.

css/properties.schema.json
skyclouds2001: It just adjusted the order according to the alphabet

dipikabh: Ah, got it, thanks for clarifying

dipikabh (approved):
Thank you so much!

PR #957, opened 2025-03-28
feat(css): Add data for `baseline-shift`

skyclouds2001:
### Description

`top | center | bottom` seems not be supported as valid css property value (for chrome, firefox & webkit), so it is ignored; 

### Additional details

https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/baseline-shift
https://www.w3.org/TR/css-inline-3/#propdef-baseline-shift
https://svgwg.org/svg2-draft/text.html#BaselineShiftProperty


css/properties.json
```diff
@@ -2368,6 +2368,21 @@
     "status": "standard",
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/background-size"
   },
+  "baseline-shift": {
+    "syntax": "<length-percentage> | sub | super | top | center | bottom | baseline",
```
skyclouds2001, line 2372:
```suggestion
    "syntax": "<length-percentage> | sub | super | baseline",
```

pepelsbey (commented):
All IDs are in lower camelcase, I wonder if we should keep it this way.

Otherwise, looks good.

pepelsbey (approved):
Thank you!

PR #959, opened 2025-04-03
feat(css): Add data for `<text-edge>`

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/text-edge


bsmth (approved):
Thanks!

PR #960, opened 2025-04-03
feat(css): Add data for `::checkmark`

skyclouds2001:
### Description

this is supported since Chrome 134

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/::checkmark
https://drafts.csswg.org/css-forms-1/#selectordef-checkmark


bsmth (approved):
ty!

PR #961, opened 2025-04-03
feat(css): Add data for `::picker-icon`

skyclouds2001:
### Motivation

this is supported in chrome since v134

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/::picker-icon
https://drafts.csswg.org/css-forms-1/#selectordef-picker-icon


bsmth (approved):
ty!

PR #962, opened 2025-04-03
feat(css): Add data for `::picker()`

skyclouds2001:
### Description

this is supported in chrome since v134

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/::picker
https://drafts.csswg.org/css-forms-1/#selectordef-picker


pepelsbey (approved):
Looks good! Thank you :)

PR #963, opened 2025-04-03
feat(css): Add data for `stroke-color`

skyclouds2001:
### Description

currently only safari support this, since v11.1

also notice that though spec defines `<color>#`, but syntax like `red, red` is not supported at present via manually test, so `<color>` is used at present

### Additional details

https://drafts.fxtf.org/fill-stroke-3/#stroke-color


bsmth (approved):
LGTM, thanks. Appears we're missing a page for it

PR #964, opened 2025-04-03
feat(css): Add data for `scroll-initial-target`

skyclouds2001:
### Description

this is supported since chrome 133

### Additional details

https://drafts.csswg.org/css-scroll-snap-2/#propdef-scroll-initial-target


pepelsbey (approved):
Looks good! Thank you :)

PR #965, opened 2025-04-03
chore(css): Update group field for css selectors

skyclouds2001:
### Description

to make the group field really represents which module the feature is defined


css/selectors.json
```diff
@@ -146,25 +130,20 @@
   ":active-view-transition": {
```
bsmth, line 130: Commenting for traceability: https://www.w3.org/TR/css-view-transitions-2/#the-active-view-transition-pseudo

bsmth:
Thanks a lot. In general, this is technically correct, so I would be curious to get another opinion again on whether:

>make the group field really represents which module the feature is defined

is what consumers expect as opposed to these returned when someone is looking for a list of pseudo-classes or pseudo-elements. "Defined in" versus "can be classed as" is a good distinction to double-check.

bsmth (approved):
I think this is good. @pepelsbey do you agree?

pepelsbey (approved):
Yep, makes sense to me! Thank you :)

PR #966, opened 2025-04-03
chore(css): Mark `::cue-region` as experimental

skyclouds2001:
### Description

per bcd, `::cue-region` is not supported by any browser at present, so this feature should be marked as experimental


pepelsbey (approved):
Looks good! Thank you :)

PR #967, opened 2025-04-03
feat(css): Add data for `ruby-overhang`

skyclouds2001:
### Description

per bcd, it is supported in safari since v18.2

### Additional details

https://drafts.csswg.org/css-ruby/#propdef-ruby-overhang


bsmth (approved):
nice, tnx

PR #969, opened 2025-04-04
feat(css): Add data for `object-view-box`

skyclouds2001:
### Description

this is supported since chrome v102

### Additional details

https://drafts.csswg.org/css-images-5/#propdef-object-view-box


pepelsbey (approved):
Looks good! Thank you :)

PR #972, opened 2025-04-04
feat(css): Add data for `font-width`

skyclouds2001:
### Description

it is getting supported in safari at present

### Additional details

https://drafts.csswg.org/css-fonts/#propdef-font-width


bsmth:
It looks good from a data perspective. Do you have any support information to link to for Safari?

bsmth (approved):
Thanks!

PR #974, opened 2025-04-04
fix: Replace Markdown formatting with HTML

Josh-Cena:
This is showing up as raw markdown, so we should use plain HTML instead

### DIFF
--- a/l10n/css.json
+++ b/l10n/css.json
@@ -1495,7 +1495,7 @@
     "ru": "относятся к размеру маски области рисования минус размер маски слоя изображения (смотрите статью о {{cssxref(\"background-position\")}})"
   },
   "referToSVGViewportDiagonal": {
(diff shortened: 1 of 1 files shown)


l10n/css.json
```diff
@@ -1495,7 +1495,7 @@
     "ru": "относятся к размеру маски области рисования минус размер маски слоя изображения (смотрите статью о {{cssxref(\"background-position\")}})"
   },
   "referToSVGViewportDiagonal": {
-    "en-US": "refer to the normalized diagonal measure of the current SVG viewport's applied {{SVGAttr(\"viewBox\")}}, or of the viewport itself if no `viewBox` is specified"
+    "en-US": "refer to the normalized diagonal measure of the current SVG viewport's applied {{SVGAttr(\"viewBox\")}}, or of the viewport itself if no <code>viewBox</code> is specified"
```
bsmth, line 1498: Shows up here: https://developer.mozilla.org/en-US/docs/Web/CSS/stroke-dasharray

bsmth (approved):
Great, tnx 👍🏻

Josh-Cena:
I see, good to know

PR #976, opened 2025-04-05
feat(css): Update syntax for `content` property

skyclouds2001:
### Description

update `content` property: add `attr()`
update `<content-list>` syntax: add `attr()`; remove `contents` `leader()` and `<target>` which are not supported and non-standard at all

note that spec defines as `[ <string> | <image> | <attr()> | contents | <quote> | <leader()> | <target> | <string()> | <content()> | <counter> ]+`, but many are not-supported so not added at present

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/content
https://drafts.csswg.org/css-content/#propdef-content
https://drafts.csswg.org/css-content/#typedef-content-content-list

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -4492,7 +4492,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/container-type"
   },
   "content": {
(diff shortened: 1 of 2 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #977, opened 2025-04-05
feat(css): Update syntax for `@page/size` descriptor

skyclouds2001:
### Description

the change include limit to the range of `<length>`

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/@page/size
https://drafts.csswg.org/css-page/#descdef-page-size


bsmth (approved):
Thank you!

PR #978, opened 2025-04-05
feat(css): Update syntax for `@counter-style/additive-symbols` descriptor

skyclouds2001:
### Description

this change support specified range limit on `<integer>` syntax

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/@counter-style/additive-symbols


bsmth (approved):
LGTM, thanks 👍🏻

PR #979, opened 2025-04-05
feat(css): Update syntax for `@counter-style/pad` descriptor

skyclouds2001:
### Description

this change includes range limit on `<integer>`

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/@counter-style/pad


dipikabh (approved):
Thank you!

PR #982, opened 2025-04-20
feat(css): Add `anchor-center` keyword value support for `{align,justify}-{items,self}`

skyclouds2001:
### Description

the syntax update add support for the `anchor-center` keyword value

these features are support since chrome v125

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/align-items
https://developer.mozilla.org/en-US/docs/Web/CSS/align-self
https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items
https://developer.mozilla.org/en-US/docs/Web/CSS/justify-self
https://drafts.csswg.org/css-anchor-position/#valdef-justify-self-anchor-center

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -1672,7 +1672,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/align-content"
   },
   "align-items": {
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #983, opened 2025-04-20
feat(css): Update syntax for `bottom` `top` `left` `right`

skyclouds2001:
### Description

add support for `anchor()` and `anchor-size()`, which are supported since chrome v125 and chrome v132

also combine `<length>` and `<percentage>` into `<length-percentage>`

also add these properties to CSS Anchor Positioning group since this module also defines some of these features

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/bottom
https://developer.mozilla.org/en-US/docs/Web/CSS/top
https://developer.mozilla.org/en-US/docs/Web/CSS/left
https://developer.mozilla.org/en-US/docs/Web/CSS/right
https://drafts.csswg.org/css-anchor-position/#funcdef-anchor

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -3705,12 +3705,13 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/border-width"
   },
   "bottom": {
-    "syntax": "<length> | <percentage> | auto",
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #984, opened 2025-04-20
feat(css): Update syntax for `margin-*`

skyclouds2001:
### Description

the update add support for `anchor-size()` value, which are supported since chrome v132

also add these features to CSS Anchor Positioning group

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/margin-bottom
https://developer.mozilla.org/en-US/docs/Web/CSS/margin-left
https://developer.mozilla.org/en-US/docs/Web/CSS/margin-right
https://developer.mozilla.org/en-US/docs/Web/CSS/margin-top
https://drafts.csswg.org/css-anchor-position/#funcdef-anchor-size

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -6674,12 +6674,13 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/margin-block-start"
   },
   "margin-bottom": {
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #986, opened 2025-04-20
feat(css): Update syntax for `appearance`

skyclouds2001:
### Description

update syntax to match the spec, removes some unknown values (non-standard and no-support)

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/appearance
https://drafts.csswg.org/css-ui/#propdef-appearance
https://drafts.csswg.org/css-ui/#typedef-appearance-compat-auto
https://drafts.csswg.org/css-ui/#typedef-appearance-compat-special


pepelsbey (approved):
Looks good! Thank you :)

PR #987, opened 2025-04-21
feat(css): Update the syntax for `background-*`

skyclouds2001:
### Description

for `background`, simply the syntax to match the spec
for `<final-bg-layer>` & `<bg-image>`, adjust order to match the spec
for `<bg-size>`, update range value support, to be consistent with the spec

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/background
https://drafts.csswg.org/css-backgrounds/#propdef-background
https://drafts.csswg.org/css-backgrounds/#typedef-bg-layer
https://drafts.csswg.org/css-backgrounds/#typedef-final-bg-layer
https://drafts.csswg.org/css-backgrounds/#typedef-bg-image

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -2101,7 +2101,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/backface-visibility"
   },
   "background": {
-    "syntax": "[ <bg-layer> , ]* <final-bg-layer>",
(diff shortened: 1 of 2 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #988, opened 2025-04-21
feat(css): Update syntax of `mix-blend-mode`

skyclouds2001:
### Description

for `mix-blend-mode`, update syntax to support the `plus-darker` keyword value, which is supported since Safari 9

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode
https://developer.mozilla.org/en-US/docs/Web/CSS/blend-mode
https://drafts.fxtf.org/compositing/#propdef-mix-blend-mode
https://drafts.fxtf.org/compositing/#ltblendmodegt

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -7425,7 +7425,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/min-width"
   },
   "mix-blend-mode": {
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #989, opened 2025-04-24
feat(css): Update syntax for `border-*-radius`

skyclouds2001:
### Description

the change add range limit to `<length-percentage>` type

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/border-bottom-left-radius
https://developer.mozilla.org/en-US/docs/Web/CSS/border-bottom-right-radius
https://developer.mozilla.org/en-US/docs/Web/CSS/border-top-left-radius
https://developer.mozilla.org/en-US/docs/Web/CSS/border-top-right-radius
https://drafts.csswg.org/css-backgrounds/#border-radius
https://drafts.csswg.org/css-tables/#border-collapse-property

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -2725,7 +2725,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/border-bottom-color"
   },
   "border-bottom-left-radius": {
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you :)

PR #991, opened 2025-04-25
feat(css): Simplify syntax for `border-spacing`

skyclouds2001:
### Description

this update the syntax to the spec, which is also a syntax simplfy

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/border-spacing
https://drafts.csswg.org/css-tables-3/#propdef-border-spacing


pepelsbey (approved):
Looks good! Thank you :)

PR #992, opened 2025-05-05
feat(css): Update syntax for `font`

skyclouds2001:
### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/font
https://drafts.csswg.org/css-fonts/#propdef-font
https://drafts.csswg.org/css-fonts/#font-variant-css21-values
https://drafts.csswg.org/css-fonts/#system-family-name-value
https://drafts.csswg.org/css-fonts/#font-width-css3-values

Relates to https://github.com/mdn/data/issues/230

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -5200,7 +5200,7 @@
     "syntax": "<font-stretch-absolute>",
     "media": "visual",
     "inherited": true,
-    "animationType": "see",
+    "animationType": "fontStretch",
     "percentages": "no",
     "groups": [
       "CSS Fonts"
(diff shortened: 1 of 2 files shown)


css/properties.json
```diff
@@ -5200,7 +5200,7 @@
     "syntax": "<font-stretch-absolute>",
     "media": "visual",
     "inherited": true,
-    "animationType": "fontStretch",
+    "animationType": "see",
```
skyclouds2001, line 5203:
```suggestion
    "animationType": "fontStretch",
```

bsmth (approved):
Nice, thanks a lot.

PR #995, opened 2025-06-08
feat(css): Add `view-transition-name: match-element`

skyclouds2001:
### Description

add `view-transition-name: match-element`, which is supported in chrome v137

### Additional details

https://github.com/mdn/browser-compat-data/pull/26964
https://github.com/mdn/content/pull/39759
https://chromestatus.com/feature/5098745710247936
https://drafts.csswg.org/css-view-transitions-2/#auto-vt-name

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -10759,7 +10759,7 @@
     "status": "standard"
   },
   "view-transition-name": {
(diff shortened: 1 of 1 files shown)


dipikabh (approved):
Thanks a lot!

PR #999, opened 2025-07-11
Remove broken links in CSS data

Josh-Cena:
- Remove a few broken links
- `-moz-image-region` should no longer be documented; see https://github.com/mdn/content/pull/40314
- `-moz-orient`: another case where mentions of XUL should be removed

BEGIN_COMMIT_OVERRIDE
fix(css): Remove broken links
END_COMMIT_OVERRIDE


bsmth (approved):
Looking good, thanks 👍🏻

Josh-Cena:
Perhaps we should have commit linting if it's important further down

PR #1001, opened 2025-07-28
feat(css): Add appliesto definition for display:contents

kannanwisen:
### Description

This pull request adds a new key, "allElementsNoEffectIfDisplayNoneOrContents", to the properties.schema.json to more accurately describe the applicability of certain CSS properties.

The existing key "allElementsNoEffectIfDisplayNone" is insufficient for properties like float, which also have no effect on elements with display: contents.

This change allows for more precise documentation by providing a key that covers both none and contents.

### Additional details

This is achieved in three steps:

1. Updating the schema to allow a new key.
2. Adding the new key and its corresponding text definition.
3. Updating the float property in properties.json to use this new, more accurate key.

### DIFF
--- a/css/properties.schema.json
+++ b/css/properties.schema.json
@@ -202,6 +202,7 @@
         "allElementsExceptTableRowColumnGroupsTableRowsColumns",
         "allElementsExceptTableRowGroupsRowsColumnGroupsAndColumns",
         "allElementsNoEffectIfDisplayNone",
+        "allElementsNoEffectIfDisplayNoneOrContents",
         "allElementsSomeValuesNoEffectOnNonInlineElements",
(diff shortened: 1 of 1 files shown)


css/properties.schema.json
```diff
@@ -202,6 +202,7 @@
         "allElementsExceptTableRowColumnGroupsTableRowsColumns",
         "allElementsExceptTableRowGroupsRowsColumnGroupsAndColumns",
         "allElementsNoEffectIfDisplayNone",
+        "allElementsNoEffectIfDisplayNoneOrContents",
```
skyclouds2001: I didn't find any reference about this key, so why add it?

PR #1009, opened 2025-09-24
feat(css): Update syntax for `cursor`

charlessuh:
### Description

update `cursor` property: use `<cursor-predefined>`
create `<cursor-predefined>` syntax

### Motivation

make it easier to add / extend to the list

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#formal_definition
https://drafts.csswg.org/css-ui-3/#propdef-cursor
https://drafts.csswg.org/css-ui-3/#valdef-cursor-cursor-predefined

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -4557,7 +4557,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/counter-set"
   },
   "cursor": {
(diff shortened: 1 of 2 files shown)


bsmth (approved):
Sure, thanks 👍🏻

PR #1015, opened 2025-10-12
feat(css): Update `attr()` syntax to CSS Values Level 5

ryo-manba:
### Description

Update `attr()` function syntax to match CSS Values and Units Level 5 spec

### Motivation

The `attr()` syntax has been updated.

### Additional details

https://drafts.csswg.org/css-values-5/#attr-notation


pepelsbey (approved):
Looks good! Thank you :)

PR #1053, opened 2026-03-18
ci(auto-merge): sync workflow

caugner:
### Description

Syncs the `auto-merge.yml` workflow to ensure it uses the reusable workflow from `mdn/workflows` with the correct configuration:

- Sets `permissions: { contents: read }`
- Uses `mdn/workflows/.github/workflows/auto-merge.yml@main`
- Sets `auto-merge: true` only if the repository's existing workflow already auto-merges

Part of https://github.com/mdn/fred/issues/1395.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,0 +1,14 @@
+name: auto-merge
+
+on:
+  pull_request_target:
+
+permissions:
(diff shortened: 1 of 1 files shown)


caugner:
This repo only has two npm dependencies that don't get update anymore, so we don't really need an auto-merge workflow.

PR #1057, opened 2026-03-28
feat(css): add schema validation and supplementary documentation for custom properties

jonathandewitt-dev:
### Description

Adds `patternProperties` validation to the CSS properties schema to explicitly validate custom property names (CSS variables). Custom properties matching the pattern `^--` are now recognized as valid property keys in the schema.

### Motivation

The schema currently validates properties through `additionalProperties`, which treats every key individually without recognizing patterns. While `properties.json` contains a `"--*"` entry to document custom properties, the schema doesn't express that **any** key starting with `--` is valid (e.g., `--brand-color`, `--spacing-md`).

This creates several problems:

1. **Semantic mismatch**: The `"--*"` entry represents a pattern, but the schema has no pattern validation
2. **Downstream confusion**: Libraries like `csstype` must hardcode special-case handling for custom properties because the schema doesn't express the pattern (see [frenic/csstype#189](https://github.com/frenic/csstype/issues/189))
3. **Incomplete validation**: Tools consuming this schema can't programmatically validate that custom property names are structurally valid
4. **Spec alignment**: The CSS spec defines custom properties as a naming pattern, not individual properties—the schema should reflect this

Adding `patternProperties` makes the schema semantically accurate and enables automated validation of custom properties without special-case logic.

### Additional details

- Pattern uses `^--` to match the CSS Custom Properties spec ([CSS Variables Level 1](https://www.w3.org/TR/css-variables-1/#defining-variables))
- The existing `"--*"` entry in `properties.json` is preserved as documentation/reference
- Schema change is backwards compatible—only adds validation, doesn't restrict existing valid data
- Documentation updated in `css/properties.md` to explain the pattern validation

The most notable downstream consumer is frenic/csstype#63, because several large ecosystems including giants like React all depend on the `csstype` library, which itself depends on `mdn/data`. As a result, there are StackOverflow questions and GitHub issues _everywhere_ regarding this lack of schema recognition.

### DIFF
--- a/css/properties.md
+++ b/css/properties.md
@@ -117,3 +117,7 @@ There are 3 more properties that are optional:
 * `mdn_url` (string): a URL linking to the property's page on MDN. This URL must omit the localization part of the URL (such as `en-US/`).
 * `stacking` (boolean): Whether or not the property creates a stacking context. See [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) for details.
 * `alsoAppliesTo` (enum): To which elements the property also applies to. See the schema for [a list of enums](https://github.com/mdn/data/blob/main/css/properties.schema.json#L253).
+
+### Custom Properties
+
+Custom properties are a special case, since they have variable keys. These keys must be prefixed by two dashes, (e.g. `--brand-color`), and are validated by a regular expression in the [`patternProperties` section of the schema](https://github.com/mdn/data/blob/main/css/properties.schema.json#L472).
--- a/css/properties.schema.json
+++ b/css/properties.schema.json
@@ -468,5 +468,10 @@
         "$ref": "#/definitions/mdn_url"
       }
     }
+  },
+  "patternProperties": {
+    "^--": {
+      "$ref": "#/definitions/additionalProperties"
+    }
   }
 }


pepelsbey (approved):
Looks good! Thank you 🙂

caugner:
There were build failures in this PR, which causes the lint workflow to fail with this error:

```
css/properties.json
  Style – OK
/home/runner/work/data/data/node_modules/ajv/dist/compile/index.js:119
        throw e;
        ^

MissingRefError: can't resolve reference #/definitions/additionalProperties from id #
    at Object.code (/home/runner/work/data/data/node_modules/ajv/dist/vocabularies/core/ref.js:22:19)
    at Object.keywordCode (/home/runner/work/data/data/node_modules/ajv/dist/compile/validate/keyword.js:12:13)
    at /home/runner/work/data/data/node_modules/ajv/dist/compile/validate/iterate.js:16:35
    at CodeGen.code (/home/runner/work/data/data/node_modules/ajv/dist/compile/codegen/index.js:438:13)
    at CodeGen.block (/home/runner/work/data/data/node_modules/ajv/dist/compile/codegen/index.js:567:18)
    at Object.schemaKeywords (/home/runner/work/data/data/node_modules/ajv/dist/compile/validate/iterate.js:16:13)
    at typeAndKeywords (/home/runner/work/data/data/node_modules/ajv/dist/compile/validate/index.js:126:15)
    at subSchemaObjCode (/home/runner/work/data/data/node_modules/ajv/dist/compile/validate/index.js:113:5)
    at Object.subschemaCode (/home/runner/work/data/data/node_modules/ajv/dist/compile/validate/index.js:87:13)
    at Object.applySubschema (/home/runner/work/data/data/node_modules/ajv/dist/compile/subschema.js:17:16) {
  missingRef: '#/definitions/additionalProperties',
  missingSchema: ''
}
```

@jonathandewitt-dev @pepelsbey Is one of you able to look into these?

Edit: I have now made the failing check required for PR merge.

jonathandewitt-dev:
> There were build failures in this PR, which causes the lint workflow to fail

Oh no, I didn't think there would be an error for a simple schema change, I really should have checked that. Is this patched now?

jonathandewitt-dev:
@caugner I see it was not patched yet, so I just now opened a PR to fix it. https://github.com/mdn/data/pull/1065

PR #1058, opened 2026-04-03
feat(css): Update syntax and initial value for `rx` `ry`

sb3nder:
This PR updates the sintax and initial value of `rx` and `ry`.

### Additional details

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/rx
https://w3c.github.io/svgwg/svg2-draft/geometry.html#RX

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/ry
https://w3c.github.io/svgwg/svg2-draft/geometry.html#RY

https://github.com/w3c/svgwg/commit/ac28b82f983f99ce54722ac28c9f6c61544e417d

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -9364,7 +9364,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/ruby-position"
   },
   "rx": {
-    "syntax": "<length> | <percentage>",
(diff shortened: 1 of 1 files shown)


pepelsbey (approved):
Looks good! Thank you 🙂

PR #1061, opened 2026-04-10
feat(css): Add and fix `-webkit-border-start/end/before/after-*`

sb3nder:
### Description

This PR fixes `-webkit-border-before` and adds the missing `-webkit-border-*` logical properties.
It also includes the normative change to computed values for these properties. 
See: https://github.com/w3c/csswg-drafts/issues/11494


pepelsbey (approved):
Looks good! Thank you 🙂

PR #1062, opened 2026-04-11
fix(css): Fix syntax for `column-rule-style/width` and canonical order

sb3nder:
### Description

This PR fixes some typos in the `column-rule` and `column-rule-style/width` properties, and does not include any normative updates.


pepelsbey (approved):
Looks good! Thank you 🙂

PR #1065, opened 2026-04-21
fix(css): correct the malformed reference in patternProperties for custom properties schema.

jonathandewitt-dev:
### Description

Fixes the incorrect reference pattern that was breaking the lint check.

### Motivation

Fix requested by @caugner

Introduced by #1057


pepelsbey (approved):
Looks good! Thank you 🙂

PR #1070, opened 2026-05-17
Add data for border-shape property

chrisdavidmills:
### Description

Chrome adds support for the `border-shape` property; see https://chromestatus.com/feature/5459864205393920. This PR adds data for this property, but also adds the `shape()` function to the `<basic-shape>` value syntax, as it was missing.


pepelsbey (approved):
Looks good! Thank you 🙂

PR #1072, opened 2026-05-21
fix(css): `object-position` is not inherited

kane50613:
### Description

`object-position` is not inherited by default

- 4.6. Positioning Objects: the [object-position](https://drafts.csswg.org/css-images-3/#propdef-object-position) property
- WPT Inheritance of CSS Images properties: https://github.com/web-platform-tests/wpt/blob/master/css/css-images/inheritance.html

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -8339,7 +8339,7 @@
   "object-position": {
     "syntax": "<position>",
     "media": "visual",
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
Yes, this all checks out. Thanks for the fix, @kane50613!

PR #1074, opened 2026-05-22
Add data for view-transition-scope property

chrisdavidmills:
### Description

This PR adds data for the [`view-transition-scope`](https://drafts.csswg.org/css-view-transitions-2/#view-transition-scope-prop) CSS property, which is added to Chrome 147+ as part of [element-scoped view transitions](https://developer.chrome.com/docs/css-ui/view-transitions/element-scoped-view-transitions).

I am documenting this feature in https://github.com/mdn/content/pull/44217.


dipikabh (approved):
Looks good, thanks.

PR #1079, opened 2026-07-16
fix(css): Correct border-shape group

chrisdavidmills:
### Description

The `border-shape` property was put in the wrong group, and I'm wondering if this is why the formal definition isn't rendering on the page.


dipikabh (approved):
Either way, good to fix this. Thanks!
Let's see if it helps with the formal definition

PR #1087, opened 2026-08-05
feat: updated the value for `text-decoration-inset`

dletorey:
### Description

- updated the value for `text-decoration-inset` from `<length>` to `<length-percentage>`

### Motivation

- Working on [MDN issue #44840](https://github.com/mdn/content/issues/44840)

- [Content PR](https://github.com/mdn/content/pull/45028)
- [Firefox Release PR](https://github.com/mdn/content/pull/45027)

### DIFF
--- a/css/properties.json
+++ b/css/properties.json
@@ -10947,7 +10947,7 @@
     "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-decoration-color"
   },
   "text-decoration-inset": {
-    "syntax": "<length>{1,2} | auto",
+    "syntax": "<length-percentage>{1,2} | auto",
     "media": "visual",
     "inherited": false,
     "animationType": "byComputedValue",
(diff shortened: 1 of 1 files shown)


css/properties.json
dipikabh: Oddly, this entry in the [spec](https://drafts.csswg.org/css-text-decor-4/#text-decoration-inset-property) still says "N/A":

```
Percentages: N/A
```

whereas it could possibly be updated to something like `"referToElementFontSize"`.

Same for "computed". It might need to be updated to `"asSpecified"` but spec still says:

```
Computed value: specified keyword or absolute length
```

dipikabh (approved):
Your update looks good. Thank you!

css/properties.json
dletorey: This has now changed and moved on since. `referToSizeOfElement` seems to be the correct value now

dipikabh (approved):
Thanks!

PR #1089, opened 2026-08-14
feat: Add data for text-fit property

chrisdavidmills:
### Description

Chrome 150 adds support for the CSS [`text-fit`](https://drafts.csswg.org/css-text-5/#text-fit-property) property; see https://chromestatus.com/feature/5104141688635392.

This PR adds data for the new property


pepelsbey (approved):
Looks good! Thank you 🙂

PR #1093, opened 2026-08-28
feat: Mark zoom property as animatable

chrisdavidmills:
### Description

From Chrome 150 onwards, the [`zoom`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/zoom) CSS property is animatable; see https://chromestatus.com/feature/5183671737909248.

The [spec](https://drafts.csswg.org/css-viewport/#zoom-property) has been updated so that `zoom` is now marked as animated by computed value type. This PR updates the `zoom` data to reflect this.


pepelsbey (approved):
Looks good! Thank you 🙂

PR #1097, opened 2026-09-10
feat: add data for scroll-axis-lock

chrisdavidmills:
### Description

This PR adds data for the [`scroll-axis-lock`](https://drafts.csswg.org/css-overflow-5/#scroll-axis-lock) property.


dipikabh (approved):
LGTM, thank you!

PR #1099, opened 2026-09-17
feat: Add window-drag CSS property

chrisdavidmills:
### Description

This PR adds data for the [`window-drag`](https://drafts.csswg.org/css-ui-4/#window-drag) property, support for which has been added in Chrome 152 (see https://chromestatus.com/feature/5201338641285120).


dipikabh (approved):
Looks good, thank you!

